PR #18, opened 2024-03-22
chore: extract trie types to own module

rphmeier:

PR #16, opened 2024-03-21
chore: more rigorous representation

rphmeier:

PR #19, opened 2024-03-22
feat: add proof data structure

rphmeier:


rphmeier:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/19?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#19** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/19?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#18** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/18?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/17?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/16?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

pepyakin:
According to my calculations, the ratio of re-used nodes in a trie with uniformly distributed keys is less than 1% on average and from that I am wondering if we would be better off avoiding multiproofs, at least for now.

core/src/prove.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// If the `LeafData` is `Some`, this is a leaf. Otherwise, it's a terminal.
+    /// Leaves can be used either to prove the value of a key or to prove that a key doesn't exist
+    /// in the trie by contradiction. Terminals can be used to prove that a key doesn't exist in the
+    /// trie.
+    ///
+    /// These should be sorted by depth (descending) and then lexicographically (ascending) within
+    /// thesame depth.
+    pub relevant_terminals: Vec<(BitPath, Option<LeafData>)>,
+    /// All additional nodes required as part of this proof. These are ordered positionally.
+    pub siblings: Vec<Node>,
```
gabriele-0201, line 29: What do you mean by 'positionally'? Is the position related to the 'number rule'?

I see a possible implementation if the siblings are stored in order as needed, and also the terminal nodes are not provided as you described, but they should just be ordered from left to right (if you imagine the tree drawn with all branches going down). The pseudocode would be:

```pseudo

sibiling_index = 0;

for terminal_index in 0..relevant_terminals.len() - 1 {
    
    // check if the current terminal and the next one are sibiling, so last bit 
    // in the sense that both key-path should be of the same lenght and only the last bit
    // of the usefull bit in the key path should differ
    let current_node = relevant_terminals[terminal_index]

    loop {
    
        if differ_last_bit(current_node, relevant_terminals[terminal_index + 1])
            // if nodes are sibiling we can construct the next node
            // the keypath of this new internal is just the previous removed the last one
            internal = hash(current_node, relevant_terminals[terminal_index + 1])
        else if differ_last_bit(current_node, sibiling[sibiling_index]) 
            // if they are not sibiling then you take the one from the proposed sibiling.
            // as before the key path is just the current_node key_path removed the last one
            inernal = hash(current_node, sibiling[sibiling_index])
            sibiling_index += 1
        else 
            // if neither the next terminator or the current sibiling is a real sibiling then 
            // this sub-tree is finished and we can move to the next one
            // the current node will be used as sibiling as soon as all the other sub-tree reach
            // this state
            sibiling.push(internal)
        
    }
    
}

// now apply the same rule to the last terminal using all created sibiling will 
// end up in a roo
loop {

    if sibiling_index >= sibiling.len() 
        // internal is the root
        break
    if differ_last_bit(current_node, sibiling[sibiling_index]) 
        // this should always be true if the my algorithm is correct
        inernal = hash(current_node, sibiling[sibiling_index])
        sibiling_index += 1
}
```

rphmeier:
> According to my calculations, the ratio of re-used nodes in a trie with uniformly distributed keys is less than 1% on average and from that I am wondering if we would be better off avoiding multiproofs, at least for now.

I looked back over your calculation and I don't think it takes into account a "tractable representation" with leaves and terminators. I expect that improves the space savings substantially. There are also two ways a node can be "shared", leading to different space savings.
  1. Shared via same child: space is saved by not repeating the node
  2. Shared via sibling: space is saved by omitting the node entirely in the proof

One of the features we are going to ship is a multi-update proof, which requires the multi-get proof to execute.

core/src/prove.rs
gabriele-0201: > positionally here means "ordered by depth and left-to-right" effectively

Okay! It makes more sense now! But implementing an algorithm that creates or verifies a proof given this structure is not obvious to me, I will think about it!

rphmeier: I have some notes. Let me share.
The algorithm is based off of interleaving 3 lists: the terminals, all nodes generated during the verification, and the siblings.

1. instantiate an empty list "implicit_nodes: Vec<(path, node)>"
2. from d (lowest) to 0:
3. pop next terminal with depth d or implicit node with depth d 
4. if the next in implicit or terminals is the sibling, pop that
5. otherwise, set sibling = pop from `siblings`. if `nodes` empty, fail.
6. let new_node = h(node, sibling). push onto end of `implicit_nodes`
7. repeat until all empty
8. check root hash matches.
// complexity: O(logn) in the size of the trie.

pepyakin:
Yes, like I said: I was counting the re-used nodes between different keys. That is, if you insert N keys, the number of nodes that are shared between more than one path converges to around 0.61%, implying that that many nodes on average would be non-unique in proofs. That means you can squash those, although, not completely because they would need to appear in the proof at least once. If I understand correctly, this falls into your first category.

The second category was indeed not accounted in the number I provided. It allows you to eliminate an entire subpath, sure.  However, intuitively, it feels to me that the savings would not be that big. Why? As you descend by a given path, the probability you would share the a note dwindles very fast, for the layer d it would be 0.5<sup>d</sup>. Am I missing something?

If I were to bet, I'd say the savings would still hover in vicinity of 1%.

> One of the features we are going to ship is a multi-update proof, which requires the multi-get proof to execute.

Both of those could be single-point proofs in coats.

Don't get me wrong, it's not like I am against multiproofs. It just may seem better suited for the later stages of the project. After all, 1% may be a lot esp. in the context of proving.

rphmeier:
Okay, I will table this and go for single key proofs / verifications.

PR #22, opened 2024-03-23
feat: merkle proving and verification

rphmeier:

PR #23, opened 2024-03-23
feat: verified path proof functions

rphmeier:


rphmeier:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/23?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#24** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/24?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/23?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#22** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/22?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

core/src/proof.rs
```diff
@@ -100,6 +100,45 @@ pub struct VerifiedPathProof {
...
+
+    /// Get the proven path.
+    pub fn path(&self) -> &BitSlice<u8, Msb0> {
+        &self.key_path.view_bits()[..self.depth as usize]
+    }
+
+    /// Check whether this path resolves to the given leaf.
+    ///
+    /// A return value of `true` confirms that the key indeed has no value in the trie.
+    /// A return value of `false` means that either the key provided is out of scope for this path
```
rphmeier, line 117: review: a few options for "out of scope" queries in this function and `confirm_nonexistence`

1. Just return false (approach used here). pro: simple API, con: yields false negatives, may be confusing
2. Return an option. pro: clear semantics, con: complex API
3. panic. pro: simple API, con: requires checking

thoughts?

core/src/proof.rs
```diff
@@ -100,6 +100,45 @@ pub struct VerifiedPathProof {
     terminal: Option<LeafData>,
 }
 
+impl VerifiedPathProof {
+    /// Get the terminal node.
```
pepyakin, line 104: Feels weird to say that it gets you the terminal node while returning None.

core/src/proof.rs
```diff
@@ -100,6 +100,45 @@ pub struct VerifiedPathProof {
...
+    pub fn terminal(&self) -> Option<&LeafData> {
+        self.terminal.as_ref()
+    }
+
+    /// Get the proven path.
+    pub fn path(&self) -> &BitSlice<u8, Msb0> {
+        &self.key_path.view_bits()[..self.depth as usize]
```
pepyakin, line 111: Maybe warrants a comment on `depth` declaration that it's length ≤ key_path.len()

core/src/proof.rs
```diff
@@ -100,6 +100,45 @@ pub struct VerifiedPathProof {
...
+    }
+
+    /// Check whether this path resolves to the given leaf.
+    ///
+    /// A return value of `true` confirms that the key indeed has no value in the trie.
```
pepyakin, line 116: Copy paste from below.

core/src/proof.rs
pepyakin: The error can raise in two cases:

1. somebody is tricking so the execution should not proceed. A `false` is ok in this case.
2. somebody is trying to debug this piece of software or a piece of software that provides proofs. In this case, they would probably appreciate the differentiation between different cases: not a leaf but terminator, leaf but with different key path, leaf but with different value hash.

I am pretty sure that the guys from the second category (maybe us) will appreciate a more detail error description.

core/src/proof.rs
```diff
@@ -100,6 +100,45 @@ pub struct VerifiedPathProof {
...
+    }
+
+    /// Check whether this proves that a key has no value in the trie.
+    ///
+    /// A return value of `true` confirms that the key indeed has no value in the trie.
+    /// A return value of `false` means that either the provided key is out of scope or it has
+    /// a value.
+    pub fn confirm_nonexistence(&self, key_path: &KeyPath) -> bool {
+        let this_path = self.path();
+        let other_path = &key_path.view_bits::<Msb0>()[..self.depth as usize];
```
pepyakin, line 130: untrusted input requires some care.

pepyakin (approved):

gabriele-0201 (approved):

core/src/proof.rs
rphmeier: `depth` is a u8, `key_path` has 256 bits, it's impossible to fail, no?

core/src/proof.rs
rphmeier: That's protected by the type `u8`

core/src/proof.rs
rphmeier: Fair. I added more docs but will punt on adding an `enum TerminalNode`

PR #28, opened 2024-03-26
page_id: KeyPath to PageId Iterator

gabriele-0201:


gabriele-0201:
* **#28** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/28?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

core/src/pageId.rs
```diff
@@ -0,0 +1,131 @@
...
+        let sextet = self
+            .key_path
+            .view_bits::<Msb0>()
+            .chunks_exact(DEPTH)
+            .nth(self.level as usize)?;
+
+        //#[cfg(feature = "std")]
+        //println!("level: {}, sextet: {:?}", self.level, sextet);
+
+        let next_page_id: &mut BitSlice<u8, Msb0> = self.next_page_id.view_bits_mut();
```
gabriele-0201, line 59: Using Msb0 for the KeyPath is perfect, to maintain consistency I tried to use the same convention for the PageId as well. It's feasible but the indices used are not very convenient in this case, especially when working mainly with the least significant byte. Using an Lsb0 convention could simplify things.

The handling of PageId is mainly restricted to this Iterator, so there are no other places where one must remember that the PageId is a little-endian integer. What do you think? @rphmeier @pepyakin should we keep Msb0 also here?

rphmeier: How does Lsb0 make it easier?

core/src/pageId.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        let prev_page_id = self.next_page_id;
+
+        // The last bits will not be used as a sextet to identify a page.
+        // If sextets are finished, then there are no more pages to load.
+        let sextet = self
+            .key_path
+            .view_bits::<Msb0>()
+            .chunks_exact(DEPTH)
+            .nth(self.level as usize)?;
```
rphmeier, line 54: isn't this causing `O(n^2)` behavior?

rphmeier: i recommend keeping either an iterator or an index inside of the `PageIdsIterator` struct so it's not being recomputed each time

core/src/lib.rs
```diff
@@ -10,5 +10,7 @@
 
 extern crate alloc;
 
+pub mod page;
+pub mod pageId;
```
rphmeier, line 14: recommend combining the page / page ID into one `page` module. if not, this should be `page_id` - Rust is always `snake_case`

rphmeier (commented):
I'm somewhat tempted to say "use a wide int library" for this binary logic, but it may be overkill, unless there's a very light one, given that you've already implemented the logic by hand.

core/src/pageId.rs
```diff
@@ -0,0 +1,131 @@
...
+///
+/// Given a KeyPath, it will be divided into chunks of 6 bits (sextet), forming
+/// an array of sextets. Each PageId will be constructed in the following manner:
+///
+/// page_ids[i] = (prev_page_id << d) + int(sextets[i]) + 1
+///
+/// The following struct implements an Iterator of PageIds over a KeyPath
+/// to lazily construct them as needed
+struct PageIdsIterator {
+    key_path: KeyPath,
```
rphmeier, line 28: does the iterator need to own the path of just borrow it?

core/src/page.rs
```diff
@@ -0,0 +1,9 @@
+//! TODO
+
+/// Depth of the rootless sub-binary tree stored in a page
+pub const DEPTH: usize = 6;
+
+// Total number of nodes stored in one Page. It depends on the `DEPTH`
+// of the rootless sub-binary tree stored in a page following this formula:
+// (2^DEPTH) - 2
+pub const NODES_PER_PAGE: usize = 126;
```
rphmeier, line 9: could also be specified as `(1 << DEPTH) - 2`

core/src/pageId.rs
```diff
@@ -0,0 +1,131 @@
...
+            .view_bits::<Msb0>()
+            .chunks_exact(DEPTH)
+            .nth(self.level as usize)?;
+
+        //#[cfg(feature = "std")]
+        //println!("level: {}, sextet: {:?}", self.level, sextet);
+
+        let next_page_id: &mut BitSlice<u8, Msb0> = self.next_page_id.view_bits_mut();
+
+        if sextet != bits!(1, 1, 1, 1, 1, 1) {
```
rphmeier, line 61: I recommend extracting out a few standalone functions:

```rust
pub fn child_page_id(page_id, child_index) -> PageId;
pub fn parent_page_id(page_id) -> PageId;
```

and then implementing this iterator based on `child_page_id`.

rphmeier: https://github.com/thrumdev/nomt/pull/29 needs these

core/src/pageId.rs
gabriele-0201: Yes! keeping a reference to the key path and iterating it when needed is the idea, I just made the easiest solution to begin with to focus on the PageIds generation. I will implement it tomorrow!

core/src/lib.rs
gabriele-0201: Oh, yes, I'm stupid 😅

core/src/pageId.rs
gabriele-0201: Currently, it borrows it

core/src/page_id.rs
```diff
@@ -0,0 +1,170 @@
...
+impl<'a> Iterator for PageIdsIterator<'a> {
+    type Item = PageId;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let new_page_id = match self.page_id {
+            None => Some(ROOT_PAGE_ID),
+            Some(prev_page_id) => {
+                // If sextets are finished, then there are no more pages to load
+                let child_index = self.sextets.next()?.load_be::<u8>();
+                // TODO: could be nice to modify prev_page_id in place
```
rphmeier, line 96: let's remove this TODO in a follow up. Very premature optimization

core/src/page_id.rs
```diff
@@ -0,0 +1,170 @@
...
+///
+/// The following struct implements an Iterator of PageIds over a KeyPath
+/// to lazily construct them as needed
+struct PageIdsIterator<'a> {
+    sextets: ChunksExact<'a, u8, Msb0>,
+    page_id: Option<PageId>,
+}
+
+impl<'a> PageIdsIterator<'a> {
+    fn new(key_path: &'a KeyPath) -> Self {
```
rphmeier, line 78: unused - should be ignored or made public

core/src/page_id.rs
```diff
@@ -0,0 +1,170 @@
...
+/// bytes will be treated as a 256-bit big-endian integer.
+///
+/// Given a KeyPath, it will be divided into chunks of 6 bits (sextet), forming
+/// an array of sextets. Each PageId will be constructed in the following manner:
+///
+/// page_ids[i] = (prev_page_id << d) + int(sextets[i]) + 1
+///
+/// The following struct implements an Iterator of PageIds over a KeyPath
+/// to lazily construct them as needed
+struct PageIdsIterator<'a> {
```
rphmeier, line 72: pub

core/src/page_id.rs
```diff
@@ -0,0 +1,170 @@
...
+
+const MAX_CHILD_INDEX: u8 = (1 << DEPTH) - 1;
+
+pub fn child_page_id(mut page_id: PageId, child_index: u8) -> PageId {
+    // TODO: proper error handling
+    assert!(child_index & 0b11000000 == 0);
+
+    let page_id_bits = page_id.view_bits_mut::<Msb0>();
+
+    if child_index ^ MAX_CHILD_INDEX == 0 {
```
rphmeier, line 18: why not just use `==`? seems unnecessarily fancy

core/src/page_id.rs
```diff
@@ -0,0 +1,170 @@
...
+
+pub type PageId = [u8; 32];
+
+pub const ROOT_PAGE_ID: [u8; 32] = [0; 32];
+
+const MAX_CHILD_INDEX: u8 = (1 << DEPTH) - 1;
+
+pub fn child_page_id(mut page_id: PageId, child_index: u8) -> PageId {
+    // TODO: proper error handling
+    assert!(child_index & 0b11000000 == 0);
```
rphmeier, line 14: this is less readable than `assert!(child_index < MAX_CHILD_INDEX)`. We should not be making the reader spend cycles on things that can be easy unless there is a driving performance need.

rphmeier: also, generally I consider `assert` or panicking bad practice over error handling, but there is a TODO which should include the two failure cases:
  1. child index out of bounds
  2. page id overflow

core/src/page.rs
```diff
@@ -0,0 +1,8 @@
+//! TODO: add docs
+
+/// Depth of the rootless sub-binary tree stored in a page
+pub const DEPTH: usize = 6;
+
+// Total number of nodes stored in one Page. It depends on the `DEPTH`
+// of the rootless sub-binary tree stored in a page
+pub const NODES_PER_PAGE: usize = (1 << DEPTH) - 2;
```
rphmeier, line 8: (1 << DEPTH + 1) - 2 actually (my bad earlier)

core/src/page_id.rs
```diff
@@ -0,0 +1,170 @@
...
+                Some(child_page_id(prev_page_id, child_index))
+            }
+        };
+        self.page_id = new_page_id;
+        new_page_id
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
rphmeier, line 107: let's add more comprehensive tests including the overflow / out of bounds cases

rphmeier (commented):
documentation on all public items is missing, let's add that & address other comments in a follow up.

core/src/page_id.rs
gabriele-0201: Totally true, I started thinking about using a mask and I went in that direction but yes,`==` does the same thing and it is cleaner

core/src/page_id.rs
gabriele-0201: thanks, yes `assert!(child_index <= MAX_CHILD_INDEX)` is for sure more readable. I will implement proper error handling though!

PR #29, opened 2024-03-26
feat: page, pageset, and pagecursor

rphmeier:


rphmeier:
* **#29** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/29?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

core/src/page.rs
```diff
@@ -0,0 +1,379 @@
...
+        let i = (page_path.load_be::<u8>() as usize) << 1;
+        let base = (1 << (len + 1)) - 2;
+
+        (base + i, base + i + 1)
+    }
+}
+
+fn parent_page_id(_page_id: PageId) -> PageId {
+    // TODO
+    unimplemented!()
```
rphmeier, line 282: depends on #28

core/src/page.rs
```diff
@@ -0,0 +1,379 @@
...
+}
+
+fn child_page_id(_page_id: PageId, _child_idx: u8) -> PageId {
```
rphmeier, line 285: the real version of this function probably needs to fail when out of bounds.

core/src/page.rs
```diff
@@ -0,0 +1,379 @@
...
+fn node_index(page_path: &BitSlice<u8, Msb0>) -> usize {
+    let len = page_path.len();
+    if len == 0 {
+        0
+    } else if len > DEPTH {
+        node_index(&page_path[..DEPTH])
+    } else {
+        let i = page_path.load_be::<u8>() as usize;
+        let base = (1 << len) - 2;
+        base + i
```
gabriele-0201, line 258: I would move the `-2` here to emphasize that it is the 'conversion' between the number rule in the binary tree and how they are indexed in memory. I had to spend some time before figuring it out

core/src/page.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+}
+
+// Transform a bit-path to the indices of the two child positions in a page.
+//
+// The expected length of the page path is between 0 and `DEPTH - 1`, inclusive. All bits beyond
+// `DEPTH - 1` are ignored.
+fn child_node_indices(page_path: &BitSlice<u8, Msb0>) -> (usize, usize) {
```
gabriele-0201, line 266: Why is this function not just?

```rust
let i = node_index(page_path) * 2;
(i, i+1)
```

core/src/page.rs
```diff
@@ -0,0 +1,379 @@
...
+    let len = page_path.len();
+    if len == 0 {
+        (0, 1)
+    } else if len > DEPTH - 1 {
+        child_node_indices(&page_path[..DEPTH - 1])
+    } else {
+        let i = (page_path.load_be::<u8>() as usize) << 1;
```
gabriele-0201, line 273: I can't understand this shift, could you explain it?

core/src/page.rs
rphmeier: that logic would place multiple nodes at index 0

core/src/page.rs
rphmeier: It's just `* 2`, will modify for readability

core/src/page.rs
```diff
@@ -0,0 +1,379 @@
...
+/// The root page ID is fully zeroed.
+pub const ROOT_PAGE_ID: PageId = [0u8; 32];
+
+/// A read-only, borrowed page.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Page<'a> {
+    /// The underlying slice of nodes in memory.
+    /// This is expected to have a length equal to a power of two.
+    ///
+    /// The first `len - 2` slots are reserved for nodes. The last slot is the page ID.
```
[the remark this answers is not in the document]
rphmeier: Yeah, it's a backend thing which is used to resolve collisions in an on-disk hash map.

We could separate it (and indeed my first draft did exactly as you describe). I am about even on whether any approach is better. On one hand we know that the page layout in the real backend will include the page ID. On the other, that information isn't strictly needed at this level. May change,

but I will not block this PR on that refactoring.

core/src/page.rs
rphmeier: It's different than my mental model, which is that there is an offset for each length and an offset within each length - `base + i`, where `base = 2^len - 2` and `i` is the bit-string interpreted as a number.

PR #32, opened 2024-03-27
refactor: enable 'static pageset

rphmeier:
This addresses review on #29 that the `Page<'a>` struct having a lifetime is too unwieldy.

There are 4 approaches I considered to working around this:
  1. Change `Page` to own the data. This is a non-starter as it inhibits any kind of intelligent arena allocation.
  2. Change `PageSet` to be generic over a `Page: Borrow<RawPage>`. This introduces a generic which is somewhat unwieldy but also allows high flexibility. This approach is taken in this PR.
  3. Make `PageSet` a trait rather than a struct. I actually quite like this approach, as it pushes indexing and managing tree relationships between pages up the stack while limiting the amount of generics: just write `impl PageSet` in most places and you're good to go. I may revisit and do this instead later. (https://github.com/thrumdev/nomt/pull/33)
  4. Make `Page` a concrete type that references into an arena directly, with ref-counted mechanics. I didn't like this approach as the page arena is a backend responsibility.

### DIFF
--- a/core/src/page.rs
+++ b/core/src/page.rs
@@ -26,6 +26,7 @@ use crate::trie::{KeyPath, Node};
 
 use alloc::collections::BTreeMap;
 use bitvec::prelude::*;
+use core::borrow::Borrow;
 
 /// Depth of the rootless sub-binary tree stored in a page
 pub const DEPTH: usize = 6;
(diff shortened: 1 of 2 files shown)


rphmeier:
* **#32** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/32?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

rphmeier:
### Merge activity

* **Mar 28, 1:43 PM EDT**: @rphmeier started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/32).

PR #34, opened 2024-03-28
fixes: page_id and page

gabriele-0201:

PR #35, opened 2024-03-28
page_id: error handling in child_page_id

gabriele-0201:
Inside `traverse_children`, I used `unwrap` for one main logical reason: the operation of traversing the tree operated by the Cursor must be infallible given the information of the KeyPath. A KeyPath is an arbitrary string of 256 bits, so there should be no way to pass a wrong KeyPath that could lead to a wrong construction of the child index pased to `child_page_id`. 

Despite that, I still think it's valuable to return an error on `child_page_id`, mostly to detect possible bugs introduced during development. 

What do you think? Do you think instead it should be handled and `traverse_children` should return a more general error like `TraverseError`, not only covering `MissingPage` but also `ChildPageIdError`?

### DIFF
--- a/core/src/page.rs
+++ b/core/src/page.rs
@@ -223,7 +223,9 @@ impl<'a, P: PageSet + 'a> PageSetCursor<'a, P> {
                 CursorLocation::Root => ROOT_PAGE_ID,
                 CursorLocation::Page(p) => {
                     let child_page_idx = last_page_path(&self.path, self.depth).load_be::<u8>();
-                    child_page_id(*p.id(), child_page_idx)
+                    child_page_id(*p.id(), child_page_idx).expect(
+                        "Child index is 6 bits and Pages do not go deeper than the maximum layer, 42"
+                    )
                 }
             };
 
--- a/core/src/page_id.rs
+++ b/core/src/page_id.rs
@@ -28,14 +28,67 @@ pub type PageId = [u8; 32];
 /// It has an ID consisting of all zeros.
 pub const ROOT_PAGE_ID: [u8; 32] = [0; 32];
 
+/// The PageId of the leftmost page of the last layer of the page tree.
+/// It is the lowest PageId beyond which all pages with a PageId equal to
+/// or higher overflow if a child is attempted to be accessed
+pub const LOWEST_42: [u8; 32] = [
+    0, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65,
+    4, 16, 65, 4, 16, 65,
+];
+
 const MAX_CHILD_INDEX: u8 = (1 << DEPTH) - 1;
 
 /// Construct the Child PageId given the previous PageId and the child index.
 ///
 /// Child index must be a 6 bit integer, two most significant bits must be zero.
-pub fn child_page_id(mut page_id: PageId, child_index: u8) -> PageId {
-    // TODO: proper error handling
-    assert!(child_index <= MAX_CHILD_INDEX);
+/// Passed PageId must be a valid PageId and be located in a layer below 42 otherwise
+/// `PageIdOverflow` will be returned.
+pub fn child_page_id(mut page_id: PageId, child_index: u8) -> Result<PageId, ChildPageIdError> {
+    if child_index > MAX_CHILD_INDEX {
+        return Err(ChildPageIdError::InvalidChildIndex);
+    }
+
+    // RootPageId is at layer zero, the second layer has index one
+    // and so on. There are 43 layers.
+    //
+    // For each layer between 1 and 42 the lowest id for layer I is:
+    //
+    // L(I) = sum_i(2^(6 * i)) where i goes from 0 to I - 1
+    //
+    // while the highest id for each layer is:
+    //
+    // H(I) = L(I) + 2^(6 * I) - 1 = sum_i(2^(6 * i)) where i goes from 1 to I
+    //
+    // So for each layer bigger than 0 the range of PageIds are:
+    //
+    // [L(I)..=H(I)]
+    //
+    // And L(I) = H(I - 1) + 1
+    //
+    // Any PageId larger than or equal to L(42) will overflow.
+    // L(42) and H(41) have the same most significant bit
+    // set to one at position 246, resulting in 9 leading zeros.
+    // If the PageId passed as an argument has fewer than 9 leading zeros,
+    // it will definitely overflow. If it has exactly 9 leading zeros,
+    // it needs to be compared to L(42). If it is higher than L(42),
+    // it will overflow.
+    let leading_zeros = page_id.view_bits::<Msb0>().leading_zeros();
+    if leading_zeros < 9 {
+        return Err(ChildPageIdError::PageIdOverflow);
+    } else if leading_zeros == 9 {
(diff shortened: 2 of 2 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/35?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#35** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/35?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#34** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/34?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

core/src/page.rs
```diff
@@ -202,7 +202,8 @@ impl<'a> PageSetCursor<'a> {
                 CursorLocation::Root => ROOT_PAGE_ID,
                 CursorLocation::Page(p) => {
                     let child_page_idx = last_page_path(&self.path, self.depth).load_be::<u8>();
-                    child_page_id(*p.id(), child_page_idx)
+                    // extracted child_page_idx will never be bigger than 6 bits
+                    child_page_id(*p.id(), child_page_idx).unwrap()
```
rphmeier, line 206: unwrap should never appear in production code if we can help it. use `expect` with a proof

core/src/page_id.rs
```diff
@@ -13,13 +30,28 @@ pub const ROOT_PAGE_ID: [u8; 32] = [0; 32];
...
 
-    if child_index ^ MAX_CHILD_INDEX == 0 {
+    // The PageId of the first page for a layer I is:
+    // sum_i(2^(6 * i))
+    // where i goes from 0 to I - 1
+    //
+    // So the first page in the last layer has in position 246
+    // the last bit set to one. Every PageId with a 1 after that
+    // position will lead to a PageOverflow.
+    if dbg!(page_id_bits.leading_zeros()) <= 9 {
```
rphmeier, line 50: stray `dbg!`

rphmeier, line 50: comment does not explain the 9 - can't say I get it.
The comment here could be improved, and the logic may be wrong as well.

I see something like an argument that:
  1. The lowest page ID at the bottom of the page tree has ID `sum_i(2^(6 * i))`
  2. Therefore any page with an ID `>=` than this value cannot have children

But that is not what is implemented. Also, unless I am mistaken the lowest page ID at the bottom of the page tree equals `sum(2^(6k) for k 0..=42)` and this has bit 252=1.

What seems to be implemented is actually a different argument:
1. The highest page ID at the 2nd to last layer is equal to `sum(2^(1+6k) for k 1..=41` and this has bit 247 set as a result.
2. Any page ID which has a bit higher than 247 set is definitely in the last layer.
3. As above, the smallest last-layer page ID starts with bit 252 set, so this is exhaustive

But this gives me the check `leading_zeros < 9`, not `<= 9` because `= 9` means that bit 247 is set. So is the code here inaccurate or am I missing something?

In any case, we could also just as easily simplify the logic and have `leading_zeros < 4` as this means that bit 252 is set.

The comment gave me none of this. It could have been much more helpful.

rphmeier:
```rust
    #[test]
    fn last_page_id_test() {
        let levels = 42;

        // min page
        let mut page_id = ROOT_PAGE_ID;

        for i in 0..42 {
            page_id = child_page_id(page_id, 0).unwrap();
        }

        // max page
        let mut page_id = ROOT_PAGE_ID;

        for i in 0..42 {
            page_id = child_page_id(page_id, 1 << 6).unwrap()
        }
    }
```

fails against this diff, btw

gabriele-0201: 😓

rphmeier:
> Despite that, I still think it's valuable to return an error on child_page_id, mostly to detect possible bugs introduced during development.

Yes, agreed. also for separation of concerns - we can handle impossible cases higher up.

core/src/page_id.rs
gabriele-0201: Sorry for the poor explanation.

You're totally right on how the layer before the last has the same bit in position 246 set to one, thanks for letting me notice that!

I mixed a couple of basic binary tree logic with the 64-ary tree we are dealing with. Within a layer, the most significant bit set to one can change (this cannot happen in a full binary tree), so working only with leading zeros is not enough.

However, `lowest(n) = sum_i(2^(1 + 6k)) for k 0..=n` doesn't appear correct to me. For example, `lowest(1)` would be `sum_i(2^(1 + 6k)) = 2 + 2^7 != 1`. Instead, what I came up with is:
image

The formula does not work for layer 0, only between 1 and 42.

Another important thing to notice is that the page tree is made up of 43 layers, not 42. This is because there are 42 child indices for each KeyPath, so starting from layer zero (root) means arriving at layer 42.

So, given 42 as the index of the last layer of the page tree, the most significant bit set to one in layer 42 is `6 * 41 = 246`.

As you described, `lowest(42) = highest(41) + 1`, so bit 246 is set to one in both, and the solution I provided cannot be correct.

>The best actual solution I can think of is to reject every value with 9 leading zeroes or fewer unless it is bitwise less than `lowest(41)`.

I like this proposed solution, with a small variation. We know that the most significant bit set to one in `highest(42) = lowest(43) - 1` (even if `lowest(43)` doesn't really exist) is `252`. So, if there are less than or equal to 4 leading zeroes, there will be a direct page overflow; if there are between 5 and 9 leading zeroes, bitwise comparison to `lowest(42)` is necessary.

rphmeier: > Another important thing to notice is that the page tree is made up of 43 layers, not 42. This is because there are 42 child indices for each KeyPath, so starting from layer zero (root) means arriving at layer 42.

sure, this depends how you look at it. I was counting from 0 as the layer below the root. There are 42 layers beyond the root which gives you a max index of 41 in that scheme. It' s a matter of definitions rather than truth.

> For example, lowest(1) would be sum_i(2^(1 + 6k)) = 2 + 2^7 != 1.

Yeah, I had this wrong. So I think the general function we are looking for is 

`path_n(depth) = sum_i(n * 64^i) i in 0..depth` 

I'll use your 1-counting which treats the depth zero as root.
This gives path_n(0) = 0 for all n.
path_1(1) = 1 * 1
path_64(1) = 64 * 1

so on.

> I like this proposed solution, with a small variation. We know that the most significant bit set to one in highest(42) = lowest(43) - 1 (even if lowest(43) doesn't really exist) is 252. So, if there are less than or equal to 3 leading zeroes, there will be a direct page overflow; if there are between 4 and 9 leading zeroes, bitwise comparison to lowest(42) is necessary.

This is just overcomplicating the solution. If a value has fewer leading zeroes than `highest(41)` then it definitely is in layer 42 and would overflow. If it has the same amount of leading zeroes it may also be in layer 2 and needs to be lexicographically compared. I don't see any benefit to adding another branch for 0-3 leading zeroes, except that we will overflow the container and this is different. But page IDs are logically 250 bits and we don't need to differentiate these cases.

gabriele-0201: > This is just overcomplicating the solution

totally agree, thinking without coding made me overthinking about the problem

rphmeier (approved):

PR #84, opened 2024-04-09
feat: store leaf data in pages alongside nodes

rphmeier:
This PR places leaf data in the pages, alongside the nodes themselves.

This enables leaves to be looked up alongside nodes, without requiring an additional lookup of the
leaf node preimage in a secondary index. 

The "child" positions for a node within a page are always empty when the node is a leaf. This
enables us to write the leaf node preimage (64 bytes) to those two 32 byte slots.

There is some additional complexity as a result. When we write a leaf node in a page, now,
we need the leaf node data alongside it. Likewise, when we relocate a leaf node we also need
to relocate the child data. This relocation happens up to once per updated leaf terminal, except when we delete nodes, so we can reserve
a kind of warm-up for the deletion case alone.


rphmeier:
* **#84** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/84?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

rphmeier:
(currently pessimizes, don't merge)

nomt/src/cursor.rs
```diff
@@ -229,8 +279,121 @@ impl PageCacheCursor {
...
+                page.set_node(&mut *write_pass.borrow_mut(), index, node);
+                dirtied.insert(page_id);
+            }
+        }
+    }
+
+    /// Place a leaf node at the current location.
+    pub fn place_leaf(&mut self, node: Node, leaf: trie::LeafData) {
+        if !trie::is_leaf(&node) {
+            return;
```
pepyakin, line 392: why not panic? shouldn't we try to make the system as tight as possible, and therefore take the every opportunity to make the fuzz or test fail?

nomt/src/cursor.rs
```diff
@@ -229,8 +279,121 @@ impl PageCacheCursor {
...
+        }
+
+        self.write_leaf_children(leaf.clone());
```
pepyakin, line 395: why clone?

nomt/src/cursor.rs
```diff
@@ -229,8 +279,121 @@ impl PageCacheCursor {
...
+    }
+
+    /// Place a non-leaf node at the current location.
+    pub fn place_non_leaf(&mut self, node: Node) {
+        if trie::is_leaf(&node) {
+            return;
+        }
+
+        // hack: this assumes that leaves are always explicitly deleted
+        // before being overwritten in order to maintain a consistent state.
```
pepyakin, line 364: This violates the liskov substitution principle by adding an additional precondition to this method. Therefore, either this must be removed or the contract of `Cursor::place_non_leaf` must be updated.

nomt/src/cursor.rs
```diff
@@ -229,8 +279,121 @@ impl PageCacheCursor {
...
+                let path = last_page_path(&self.path, self.depth);
+                match child_node_indices(path) {
+                    Some((left, right)) => (page, (left, right)),
+                    None => {
+                        let child_page_idx = path.load_be::<u8>();
+
+                        // unwrap: this is true even for the leaf node placement, as less than the
+                        // entire 42nd page is used with 256 bit keys.
+                        let child_page_id = page_id.child_page_id(child_page_idx).expect(
+                            "Child index is 6 bits and Pages do not go deeper than the maximum layer, 42"
```
pepyakin, line 299: Is that an adequate proof? It sounds like it doesn't go deeper than the layer 42, but `child_page_id` requires it to be strictly below 42. Ditto below and above.

nomt/src/cursor.rs
```diff
@@ -251,6 +414,77 @@ impl PageCacheCursor {
...
+                None
+            }
+            (NodeKind::Leaf, NodeKind::Terminator) => {
+                // compact: clear this node, move leaf up.
+
+                let prev = self.read_leaf_children();
+                self.write_leaf_children(Default::default());
+                self.place_non_leaf(trie::TERMINATOR);
+                self.up(1);
+                self.place_leaf(node, prev.clone());
```
pepyakin, line 453: Why clone?

nomt/src/cursor.rs
```diff
@@ -251,6 +414,77 @@ impl PageCacheCursor {
...
+
+                None
+            }
+            (NodeKind::Terminator, NodeKind::Leaf) => {
```
pepyakin, line 457: Can't we merge this arm with the previous one, by stating the both patterns in one arm?

nomt/src/cursor.rs
```diff
@@ -302,7 +536,7 @@ impl nomt_core::Cursor for PageCacheCursor {
     }
 
     fn seek(&mut self, path: KeyPath) {
-        PageCacheCursor::seek(self, path)
+        PageCacheCursor::seek(self, path, SeekMode::PathOnly);
```
pepyakin, line 539: Nit:

```suggestion
        let _ = PageCacheCursor::seek(self, path, SeekMode::PathOnly);
```

I saw this piece but I didn't see the `;` and did not realize that it returns a result now. Explicitly indicating discarding the result would be clearer.

nomt/src/cursor.rs
```diff
@@ -145,10 +168,37 @@ impl PageCacheCursor {
...
+                    let page_idx = {
+                        let this_page_idx = path.load_be::<u8>();
+                        if this_page_idx % 2 == 0 {
+                            this_page_idx + 1
+                        } else {
+                            this_page_idx - 1
+                        }
+                    };
+                    let child_page_id = id.child_page_id(page_idx).expect(
+                        "Child index is 6 bits and Pages do not go deeper than the maximum layer, 42"
```
pepyakin, line 189: Nit: this proof repeats verbatim several times. As discussed, the only big advantage of `expect` over `unwrap` is that it allows to identify the place of a panic even if the line number information mismatches the actual code and the source SHA is not known. Using the same message removes this advantage.

core/src/update.rs
```diff
@@ -177,7 +188,7 @@ pub fn build_sub_trie<H: NodeHasher>(
     skip: usize,
     leaf: Option<LeafData>,
     ops: &[(KeyPath, Option<ValueHash>)],
-    mut visit: impl FnMut(KeyPath, u8, Node),
+    mut visit: impl FnMut(KeyPath, u8, Node, Option<LeafData>),
```
pepyakin, line 191: Nit: care to update the doc?

nomt/src/store.rs
```diff
@@ -66,6 +66,7 @@ impl Store {
     }
 
     /// Load the preimage of a leaf, given its hash.
+    #[allow(unused)]
```
pepyakin, line 69: So is this going away or you have any other plans on it?

pepyakin (approved):

nomt/src/cursor.rs
rphmeier: yep, it's a bad hack. will mull over fixing it.

nomt/src/cursor.rs
rphmeier: The `self.sibling()` call is important

PR #86, opened 2024-04-09
bench: initial structure

gabriele-0201:
The following pr is very experimental. Every `unwrap`/`panic` and missing comments will be added/fixed in the following pull requests if this structure is accepted.

Things to notice are:
+ currently using `--initial size` with nomt makes it panic. Not sure if the error is due to how it is used or if there is a problem in nomt implementation.
+ using sov-db required a downgrade of `rocksdb` dependency. I will try to find another solution instead of downgrading it in nomt.
+ I'm not sure how keys should be handled... sov_db seems to re-hash keys to create separate databases where key, key_hash, and values are stored. I still need to figure out how they handle this; after that, I will be able to understand which approach is the best to use in benchmarking. For now, re-hashing of the keys is also part of the nomt benchmarking

edit.
SovDB stores three things in the main database:
+ node-key -> node
+ key-hash -> key-preimage
+ (key-preimage, version) -> value

As a result, what should be benchmarked is writing, reading, and committing key-preimage to value.

On the other hand, Nomt exposes the same functionalities (albeit with a different flavor) using key-hash to value.

This raises the question: should benchmarks cover the hashing of the key for both databases? I think that if we want to provide a fair comparison, both benchmarks should accept keys-preimage to value, and then each database will handle them in the appropriate way

### DIFF
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -1,0 +1,29 @@
+[package]
+name = "xtask"
+version = "0.1.0"
+authors.workspace = true
+homepage.workspace = true
+repository.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[dependencies]
+
+clap = { version = "4.4.8" , features = ["derive"] }
+anyhow = { version = "1.0.75" }
+
+# benchmarking
+hdrhistogram = "7.5.4"
+fxhash = "0.2.1"
+rand = "0.8.5"
+rand_pcg = "0.3.1"
+sha2 = { version = "0.10.6" }
+
+# sov-db
+sov-db = { git = "https://github.com/Sovereign-Labs/sovereign-sdk" }
+sov-schema-db = { git = "https://github.com/Sovereign-Labs/sovereign-sdk" }
+sov-prover-storage-manager = { git = "https://github.com/Sovereign-Labs/sovereign-sdk" }
+jmt = { git = "https://github.com/penumbra-zone/jmt.git", rev = "1d007e11cb68aa5ca13e9a5af4a12e6439d5f7b6" }
+
+# nomt
+nomt = { path = "../nomt" }
--- a/xtask/src/cli.rs
+++ b/xtask/src/cli.rs
@@ -1,0 +1,89 @@
+use clap::{Args, Parser, Subcommand};
+use std::fmt::Display;
+
+#[derive(Parser, Debug)]
+#[command(version, about, long_about = None)]
+pub struct Cli {
+    #[command(subcommand)]
+    pub command: Commands,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum Commands {
+    Bench(bench::Params),
+}
+
+#[derive(Debug, Clone, clap::ValueEnum)]
+pub enum Backend {
+    SovDB,
+    Nomt,
+}
+
+impl Display for Backend {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let name = match self {
+            Backend::SovDB => "sov-db",
+            Backend::Nomt => "nomt",
+        };
+        f.write_str(name)
+    }
+}
+
+#[derive(Debug, Clone, clap::ValueEnum)]
+pub enum Workload {
+    /// Performs 1 write into the storage
+    SetBalance,
+    /// Performs 2 reads and 2 writes into the storage
+    Transfer,
+}
+
+impl Display for Workload {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let name = match self {
+            Workload::Transfer => "transer",
+            Workload::SetBalance => "set-balance",
+        };
+        f.write_str(name)
+    }
+}
+
+pub mod bench {
+    use super::{Args, Backend, Workload};
+
+    #[derive(Debug, Args)]
+    pub struct Params {
+        /// Possible Backends to run benchmarks against
+        ///
+        /// Leave this flag empty to run benchmarks against all avaiable backends
+        ///
+        /// Use ',' to separate backends
+        #[clap(default_values_t = Vec::<Backend>::new(), value_delimiter = ',')]
+        #[arg(long, short)]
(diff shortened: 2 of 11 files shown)


gabriele-0201:
* **#86** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/86?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

xtask/src/sov_db.rs
```diff
@@ -0,0 +1,92 @@
...
+
+impl DB for SovDB {
+    fn apply_actions(&mut self, actions: Vec<Action>) {
+        self.state_db.inc_next_version();
+
+        // Actions are applied to jmt and then applied to the backend if committed
+        let jmt = JellyfishMerkleTree::<_, sha2::Sha256>::new(&self.state_db);
+
+        let mut preimages = vec![];
+        let mut tree_upadate = None;
```
rphmeier, line 42: typo: update

gabriele-0201:
> it would be better to keep this code separated from the main workspace and define it as a new binary

Do you mean to keep `xtask` in the same position without adding it to the workspace, right?
Even in that case, the `librocksdb-sys` version problem is not solved

``` shell
error: failed to select a version for `librocksdb-sys`.
    ... required by package `rocksdb v0.21.0`
    ... which satisfies dependency `rocksdb = "^0.21.0"` of package `sov-db v0.3.0 (https://github.com/Sovereign-Labs/sovereign-sdk#2cc0656d)`
    ... which satisfies git dependency `sov-db` of package `xtask v0.1.0 (/nomt/xtask)`
versions that meet the requirements `^0.11.0` are: 0.11.0+8.1.1

the package `librocksdb-sys` links to the native library `rocksdb`, but it conflicts with a previous package which links to `rocksdb` as well:
package `librocksdb-sys v0.16.0+8.10.0`
    ... which satisfies dependency `librocksdb-sys = "^0.16.0"` of package `rocksdb v0.22.0`
    ... which satisfies dependency `rocksdb = "^0.22.0"` of package `nomt v0.1.0 (/nomt/nomt)`
    ... which satisfies path dependency `nomt` of package `xtask v0.1.0 (/nomt/xtask)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the `links = "rocksdb"` value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.

failed to select a version for `librocksdb-sys` which could resolve this conflict 
```

That's super weird because generally, Cargo handles dependencies X and Y, which link with two different versions of Z, but this does not seem to be the case...

> what do you plan to do when running non-Rust code?

Do you mean, for example, backends that are not implemented in Rust? If that's the question, then my idea is to create another struct `NewDB` that implements the trait `DB`, where actions are converted to what is required by the backend and then call this backend providing informations and measuring it.

I will later push a huge refactor where the main concept is this: a workload is just a more or less complex series of reads and writes, and the backend's responsibility is to take those actions, convert them into the appropriate form, and then measure the time required to apply those actions.

This means that the backend is free to call another process and wait until completion (finding a way to provide the actions, and this varies from case to case)

rphmeier:
Pr

gabriele-0201:
Yes, I started with it, but I'm not attached to it. I can easily change it if you have an alternative!

The reason why I started with it is that I think we could use the same workload logic to execute profiling as well. It would be a little bit more convoluted, but in this scenario, xtask would have been the wrapper for benchmarking and profiling

xtask/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl Workload {
+    // Given the workload and its size, translate them into a
+    // vector of raw actions that need to be performed on the database
+    pub fn get_actions(&self, size: u64) -> Vec<Action> {
+        match self {
+            Workload::SetBalance => {
+                let mut accont_id: u64 = 0;
+                let writes = std::iter::from_fn(move || {
+                    accont_id += 1;
```
rphmeier, line 41: typo: account

rphmeier (approved):

xtask/src/cli.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        /// Size of the db before starting the benchmarks.
+        ///
+        /// The provided argument is the power of two exponent of the
+        /// number of elements already present in the storage.
+        ///
+        /// Leave it empty to specify an initial empty storage
+        #[clap(default_value = "0")]
+        #[arg(long, short)]
+        pub initial_size: Option<u8>,
```
pepyakin, line 82: Why option if it has a default value?

xtask/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use anyhow::Result;
+use bench::bench;
+use clap::Parser;
+use cli::{Backend, Cli, Commands, Workload};
+use nomt::NomtDB;
+use sov_db::SovDB;
+use std::rc::Rc;
+
+pub trait DB {
```
pepyakin, line 15: Nit: Just to let you know, that according to the Rust naming convention `Db` would've been a better choice.

xtask/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+clap = { version = "4.4.8" , features = ["derive"] }
+anyhow = { version = "1.0.75" }
+
+# benchmarking
+hdrhistogram = "7.5.4"
+fxhash = "0.2.1"
+rand = "0.8.5"
+rand_pcg = "0.3.1"
+sha2 = { version = "0.10.6" }
```
pepyakin, line 20: why sha2, out of curiousity?

xtask/src/cli.rs
gabriele-0201: That's an error, corrected in the follow-up PRs

xtask/Cargo.toml
gabriele-0201: No real reason, it is the one used in Sov-DB, so I stick with it for NOMT to have a comparison that leverages mostly tree things. In a future feature, the hashing algorithm used could become a parameter

PR #107, opened 2024-04-10
bench: refactor backend and workload logic

gabriele-0201:
Refactor how workloads are managed. They need to create their own
actions and use the provided backend to apply them.
The backend is responsible for translating actions to proper database
operations and measuring their timing.


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/107?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#107** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/107?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#86** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/86?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

xtask/src/backend.rs
```diff
@@ -0,0 +1,43 @@
...
+#[derive(Debug, Clone, clap::ValueEnum)]
+pub enum Backend {
+    SovDB,
+    Nomt,
+}
+
+#[derive(Clone, Debug)]
+#[allow(unused)]
+pub enum Action {
+    /// Batch of writes into the storage,
```
pepyakin, line 15: I am confused, why is it a batch? It seems to be a singular element. Doesn't seem that harmful either.

xtask/src/bench.rs
```diff
@@ -1,66 +1,33 @@
...
-pub fn bench(mut params: Params) -> Result<()> {
-    let actions = params.workload.get_actions(params.workload_size);
+pub fn bench(params: Params) -> Result<()> {
+    let workload = Workload::new(
+        params.workload.name.as_str(),
+        params.workload.size,
+        params
+            .workload
+            .initial_capacity
+            .and_then(|s| Some((s as u64) << 2))
```
pepyakin, line 11: Nothing serious, but doesn't << lead to a panic on overflow?

xtask/src/backend.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// Trait implemented by all backends who wants to be benchmarked.
+pub trait DB {
+    /// Apply the given actions to the storage, committing them
+    /// to the database at the end.
+    ///
+    /// The function also accepts an optional timer that will be used to measure
+    /// the relevant parts of the backend that effectively apply the actions.
+    fn apply_actions(&mut self, actions: Vec<Action>, timer: Option<&mut Timer>);
+}
```
pepyakin, line 30: Seems like the trait is not really necessary here, doesn't give any benefit.

xtask/src/nomt.rs
```diff
@@ -24,31 +27,40 @@ impl NomtDB {
...
+                    let key_path = sha2::Sha256::digest(key).into();
+                    let _value = match access.entry(key_path) {
+                        Entry::Occupied(o) => o.get().last_value().cloned(),
+                        Entry::Vacant(v) => {
+                            let value = session.tentative_read_slot(key_path).unwrap();
+                            v.insert(KeyReadWrite::Read(value.clone()));
+                            value
+                        }
+                    };
                 }
```
pepyakin, line 63: Where does the commit take place?

xtask/src/workload.rs
```diff
@@ -0,0 +1,94 @@
...
+        delete: u8,
+        update: u8,
+        size: u64,
+        additional_initial_capacity: u64,
+    },
+}
+
+impl Workload {
+    // Create a workload over some provided type of workload
+    pub fn new(name: &str, size: u64, additional_initial_capacity: u64) -> Result<Self> {
```
pepyakin, line 44: Nit: parse or something like that would be more suitable, because `new` gives off an impression that it is a simple constructor.

xtask/src/backend.rs
gabriele-0201: sorry, that comments shouldn't been there. This PR changed a lot of things and I forgot to remove those comments. thanks!

xtask/src/bench.rs
gabriele-0201: Yes, it could panic. I didn't think about it. Being inside the construction of benchmarking, I didn't put enough attention to panic!
It can be solved enforcing a bound with clap.
But the biggest problem here is that the shift is completely wrong... It should be (1 << s) to be like 2^s. I will correct it now!

xtask/src/backend.rs
gabriele-0201: Why? The point of the trait is to be able to execute different workloads on different backends. The trait makes it possible to loop over all specified backends and execute the workload on them without handling all backends separately

xtask/src/nomt.rs
gabriele-0201: At the end of the loop over all actions:

```rust
        let mut actual_access: Vec<_> = access.into_iter().collect();
        actual_access.sort_by_key(|(k, _)| *k);
        self.nomt.commit_and_prove(session, actual_access).unwrap();
```

xtask/src/backend.rs
pepyakin: Fair enough

rphmeier (approved):

PR #119, opened 2024-04-11
refactor: fix review from 84

rphmeier:
This addresses almost all of the comments from #84, which were largely style nits with one larger issue.

#84 contained hacky behavior, where the user of the Cursor had to explicitly clear leaves while also
being careful not to overwrite leaves while building a tree upwards. These were major undocumented
footguns.

This fixes that, by using `NODES_PER_PAGE` bits at the end of each page to keep track of which
positions are leaf data and which are nodes. When a node is written beyond the end of the leaf,
the leaf is invalidated and our bit vector lets us subsequently overwrite the leaf without clearing
the previous node.

### DIFF
--- a/core/src/cursor.rs
+++ b/core/src/cursor.rs
@@ -10,7 +10,9 @@ use crate::trie::{InternalData, KeyPath, LeafData, Node};
 /// This is not required to give results that make sense; higher level code is required to ensure
 /// that the nodes actually hash up to a common root.
 ///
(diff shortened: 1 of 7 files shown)


rphmeier:
* **#119** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/119?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

pepyakin (approved):

PR #122, opened 2024-04-12
bench: add custom workload

gabriele-0201:
from the cli, it is now possible to execute workloads like:
set_balace (heavy_write), heavy_read, heavy_update, or heavy_delete

PR #109, opened 2024-04-10
bench: add transfer workload

gabriele-0201:
example of usage:
```shell
cd xtask
cargo run -- bench -b nomt,sov-db -n transfer -s 1000 -i 10
```

PR #125, opened 2024-04-12
xtask: add profiling tool

gabriele-0201:
Add two commands (profile and exec) to xtask to enable executing `samply` over a custom workload

example usage: 

```rust
cd xtask
cargo run -- profile -b nomt -n heavy_read -s 10000 -c 10
```

### DIFF
--- a/xtask/src/profile.rs
+++ b/xtask/src/profile.rs
@@ -1,0 +1,58 @@
+use crate::{cli::profile::Params, workload};
+use anyhow::Result;
+use std::collections::VecDeque;
+use std::process::Command;
+
+pub fn profile(params: Params) -> Result<()> {
+    // check for samply to be installed
+    Command::new("sh")
+        .args(["-c", "command", "-v", "samply"])
+        .output()
+        .map_err(|_| {
+            anyhow::anyhow!("Install sampy (`cargo install samply`) to execute profiling")
+        })?;
+
+    let workload = workload::parse(
+        params.workload.name.as_str(),
+        params.workload.size,
+        params
+            .workload
+            .initial_capacity
+            .map(|s| 1u64 << s)
+            .unwrap_or(0),
+        params.workload.percentage_cold,
+    )?;
+
+    workload.init(&mut params.backend.instantiate(true));
+
+    // TODO: make cwd independent
+    let mut samply_args: VecDeque<String> = std::env::args().collect();
+    samply_args[1] = "exec".to_string();
+    samply_args.push_front("record".to_string());
+
+    let mut profiler_command = Command::new("samply").args(samply_args).spawn()?;
+    profiler_command.wait()?;
+
+    Ok(())
+}
+
+pub fn exec(params: Params) -> Result<()> {
+    // TODO: it would be super to avoid the construction
+    // of the workload in the profiler
+    // TODO: at least avoid constructing the init_actions that will not be used
+    let workload = workload::parse(
+        params.workload.name.as_str(),
+        params.workload.size,
+        params
+            .workload
+            .initial_capacity
+            .map(|s| 1u64 << s)
+            .unwrap_or(0),
+        params.workload.percentage_cold,
+    )?;
+
+    let mut backend_instance = params.backend.instantiate(false);
+
+    workload.run(&mut backend_instance, None);
+    Ok(())
+}
--- a/xtask/src/backend.rs
+++ b/xtask/src/backend.rs
@@ -33,10 +33,13 @@ impl Backend {
         vec![Backend::SovDB, Backend::Nomt]
     }
 
-    pub fn instantiate(&self) -> Box<dyn Db> {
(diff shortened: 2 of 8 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/125?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#125** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/125?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#124** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/124?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#123** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/123?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#122** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/122?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#114** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/114?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#109** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/109?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#116** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/116?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#108** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/108?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#107** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#86** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/86?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

xtask/src/profile.rs
```diff
@@ -0,0 +1,58 @@
+use crate::{cli::profile::Params, workload};
+use anyhow::Result;
+use std::collections::VecDeque;
+use std::process::Command;
+
+pub fn profile(params: Params) -> Result<()> {
+    // check for samply to be installed
+    Command::new("sh")
+        .args(["-c", "command", "-v", "samply"])
```
rphmeier, line 9: seems strange to pull samply in to this. why can't the user just run `samply` themselves?

rphmeier: this feels like feature/scope creep. unless absolutely necessary I don't think we should merge this.

gabriele-0201: Interesting points, I will respond in the reverse order:

1. The only one we need to care about profiling is nomt
Agree, I will remove the option anyway to specify other backends

2. we don't need a generalizable profiler for all DBs, just a benchmark
That's true, but the workflow logic created for benchmarks would be great to be reutilized to be able to execute tasks and profile them with samply

3. seems strange to pull samply in to this. why can't the user just run samply themselves?
Agree, it is strange, but probably only because we're doing it in Rust. If done with a simple .sh script, it would not have been so strange. Anyway, there are two reasons why I call `samply` from here: the first one is that workloads need to be set up before being executed, and it would not be great to sample also the initialization (for example, the flame chart would always be polluted by a huge amount of writes if the specified initial capacity is big). The second reason is just because I personally like one-command actions, so being able to just write `cargo run --profile ...` seems great, instead of something like `cargo run --prepare_backend ... && cargo build ... && sampl target/...`."

Finally, the most important thing we need to define is: should we have a different binary to profile things, or is it okay to include the execution of a workload inside the bench binary?

rphmeier: > That's true, but the workflow logic created for benchmarks would be great to be reutilized to be able to execute tasks and profile them with samply

This is feature creep and adds maintenance burden.

> If done with a simple .sh script, it would not have been so strange

No, it still would be strange. The user can just call `samply ./my_script.sh`.

rphmeier:
There is not enough justification to merge this at the current moment.

xtask/src/profile.rs
rphmeier: > it would not be great to sample also the initialization (for example, the flame chart would always be polluted by a huge amount of writes if the specified initial capacity is big).

This is fair, but I would instead prefer to address it by allowing the user to separate the initialization of the database from the benchmark. This is already desired, so it would be trivial to 

```
./the_binary initialize some_params
samply ./the_binary benchmark some_params
```

gabriele-0201: Super valid points, I like the opinion to avoid introducing complexity! Separate initialization from execution in the cli would be very simple!

PR #96, opened 2024-04-10
feat: avoid seeking in update

rphmeier:
Most of the time spent in `update` is spent in `seek`, where we unnecessarily traverse the pages. Given that the warmup workers have already done this traversal, we can replace all `seek`s with `jump`s and retrieve only a single page per edited terminal. I found this to be a substantial improvement in `update` performance.

PR #126, opened 2024-04-12
record path in seek

rphmeier:
We were redundantly recording paths for all encountered terminals. The `seek` function already
traversed to the node, so it's a simple matter to record all the siblings at that time.

In the future we could even record the page IDs and pages encountered during a `seek`. This would
be useful for two reasons:
  1. providing pages directly to the `Cursor` used in `update`
  2. parallel update scheduling

I also removed the `seek` function from the `Cursor` trait as it's not used in `update` since #96

### DIFF
--- a/core/src/proof.rs
+++ b/core/src/proof.rs
@@ -1,6 +1,5 @@
 //! Proving and verifying inclusion, non-inclusion, and updates to the trie.
 
-use crate::cursor::Cursor;
 use crate::trie::{
     self, InternalData, KeyPath, LeafData, Node, NodeHasher, NodeHasherExt, NodeKind, TERMINATOR,
 };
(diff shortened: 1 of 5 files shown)


rphmeier:
* **#126** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/126?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

core/src/proof.rs
```diff
@@ -160,32 +159,6 @@ impl VerifiedPathProof {
...
-            siblings.push(cursor.peek_sibling());
-        } else {
-            break;
-        }
-    }
-
-    let siblings: Vec<_> = siblings.into_iter().rev().collect();
-    (terminal, siblings)
-}
-
```
pepyakin, line 188: yay

nomt/src/cursor.rs
```diff
@@ -197,12 +222,12 @@ impl PageCacheCursor {
...
             }
-            // page is loaded, so this won't block.
+
             self.down(bit);
+            result.siblings.push(self.peek_sibling());
         }
 
-        // sanity: should be impossible not to encounter a node.
-        None
+        panic!("no terminal along path {}", dest.view_bits::<Msb0>());
```
pepyakin, line 230: 🎉

pepyakin (approved):

PR #137, opened 2024-04-16
experiment: rewrite page ID as list of child indices

rphmeier:


rphmeier:
* **#137** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/137?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

core/src/page_id.rs
```diff
@@ -83,19 +56,79 @@ impl ChildPageIndex {
...
+
+        let leading_zeros = uint.leading_zeros();
+        let bit_count = 256 - leading_zeros;
+        let sextets = (bit_count + 5) / 6;
+
+        if bit_count == 0 {
+            return Ok(ROOT_PAGE_ID);
+        }
+
+        // first we normalize the page ID: any sextet which is all zeros
```
gabriele-0201, line 77: Probably all the following algorithm could be simplified to

```rust
        let mut limbs = ArrayVec::new();

        for _ in 0..sextets - 1 {
            uint.sub_assign(Uint::<256, 4>::from(1));

            let x = uint & Uint::from(0b111111);
            limbs.push(x.to::<u8>());

            uint.shr_assign(DEPTH);
        }

        if uint.byte(0).trailing_zeros() < 6 {
            uint.sub_assign(Uint::<256, 4>::from(1));
            limbs.push(uint.byte(0));
        }

        limbs.reverse();
```
I tested it and it passes all tests, but the page_id test suite is not extensive, so it is likely incorrect and I may be missing some pieces

gabriele-0201 (commented):
I don't understand the idea behind this change set.

The previous state of page IDs was to perform operations on the PageId only on demand (requiring some complex bound checks that are generally useless because the last layers of pages will likely never be used).

However, at first glance, this makes moving between parent and child super fast but at the cost of introducing more overhead on creation because it will normalize all PageIds (created by the KeyPath, which is a hash with probably all 42 limbs required to be decoded) and later re-encode it when it's time to fetch a page from storage

core/src/page_id.rs
rphmeier: nice, this looks like it should work.

can you stack a PR on top of it?

rphmeier:
> and later re-encode it when it's time to fetch a page from storage

That we _don't_ re-encode the path is actually the primary motivation behind the change. We use the `limbs` as the key in RocksDB, not the fixed-width representation. The fixed-width representation is rarely needed.

gabriele-0201:
>We use the limbs as the key in RocksDB, not the fixed-width representation

but if you use `length_dependent_encoding` as key 

```rust
let page1 = ROOT_PAGE_ID.child_page_id(0);
let page2 = page1.child_page_id(0);
```

would page1 and page2 be the same? probably I don't know the inner behavior of ArrayVec and instead they might be treated differently

rphmeier:
Yes, exactly - they have representations that are based on the length (which RocksDB stores anyway)

Root -> `[]`
`root.child_page_id(0)` -> `[0]`
`root.child_page_id(0).child_page_id(0)` -> `[0, 0]`

So a page ID is actually made by "unpacking" the sextets of the keypath

gabriele-0201:
> they have representations that are based on the length

Oh, okay! Great! I still don't know if its faster to create all limbs initially or to do it only on-demand (this would mean having a PageId that stores both KeyPath and limbs). It is easy to implement, so I will give it a try and upload the results.

However, there is only one missing piece for me now: why does the `store` module use `page_id.length_dependent_encoding()` to write a page, while to load a page, it uses `page_id.encode()`? This should not work because the encodings are different, but tests and workloads execute without any problems

rphmeier:
> Oh, I just changed 'encode' and 'length_dependent_encoding' alternatively in the write and load pages, and the test passes anyway. We have to work on our test suite 

Ha. This is a bug, thanks for reviewing. Will push a fix! Yes, the tests pass because we never purge the cache.

rphmeier:
> I still don't know if its faster to create all limbs initially or to do it only on-demand (this would mean having a PageId that stores both KeyPath and limbs)

I don't either, to be honest. It's not our biggest issue right now - this PR isn't meant to "finalize" the implementation of the `PageId` but rather just focus on the length-dependent encoding. I could see our in-memory workload being faster with either one or the other approach, or it just not mattering at all in the scheme of things.

RocksDB is going to handle ordered data better than unordered, so I'm hoping that this changeset (and more to follow) will bring down the commit times for pages by quite a lot.

gabriele-0201 (approved):

core/src/page_id.rs
```diff
@@ -1,64 +1,37 @@
...
+//! blank bit pattern, and then repeatedly shifting it to the left by 6 bits, then adding the next
+//! child index, then adding 1. This disambiguated representation uniquely encodes all the page IDs
+//! in a fixed-width bit pattern as, essentially, a base-64 integer.
 
 use crate::{page::DEPTH, trie::KeyPath};
-use core::ops::{AddAssign, ShlAssign, ShrAssign, SubAssign};
+use arrayvec::ArrayVec;
 use ruint::Uint;
 
+const HIGHEST_ENCODED_42: Uint<256, 4> = Uint::from_be_bytes([
```
pepyakin, line 18: Needs docs

core/src/page_id.rs
```diff
@@ -1,64 +1,37 @@
...
+    16, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65, 4, 16, 65,
+    4, 16, 65, 4, 16, 64,
+]);
+
 /// A unique ID for a page.
-#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
-pub struct PageId(Uint<256, 4>);
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub struct PageId {
+    limbs: ArrayVec<u8, 42>,
```
pepyakin, line 26: Naming nit: This is literally a path as in key path.

nomt/src/store.rs
```diff
@@ -112,8 +115,12 @@ impl Transaction {
...
         let cf = self.shared.db.cf_handle(PAGES_CF).unwrap();
         match value {
-            None => self.batch.delete_cf(&cf, page_id.to_bytes().as_ref()),
-            Some(value) => self.batch.put_cf(&cf, page_id.to_bytes().as_ref(), value),
+            None => self
+                .batch
+                .delete_cf(&cf, page_id.length_dependent_encoding()),
+            Some(value) => self
+                .batch
+                .put_cf(&cf, page_id.length_dependent_encoding(), value),
```
pepyakin, line 123: Wouldn't it be more beneficial if this was length dependent **and** packed. Otherwise limb wastes 2 bits.

pepyakin (approved):

pepyakin:
### Merge activity

* **Apr 22, 10:08 AM EDT**: @pepyakin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/137).

nomt/src/store.rs
rphmeier: I am not sure how to do that right now. We would have to chop off leading bits rather than bytes.

PR #152, opened 2024-04-23
opt: async wal flush

pepyakin:
perform flushes of WAL in the background thread, reserving the main
thread for updating.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -162,6 +162,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f7144d30dcf0fafbce74250a3963025d8d52177934239851c917d29f1df280c2"
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ab3db02a9c5b5121e1e42fbdb1aeb65f5e02624cc58c43f2884c6ccac0b82f95"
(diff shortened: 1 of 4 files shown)


pepyakin:
* **#152** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/152?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

rphmeier:
I am pretty sure this is just bypassing the WAL altogether.

Running this code is much faster, but the background thread spends zero time flushing the WAL (because there is nothing to flush).

A couple sources https://github.com/facebook/rocksdb/wiki/WAL-Performance and http://rocksdb.org/blog/2017/08/25/flushwal.html say that the "correct" way to do this is to:
  1. set manual WAL flush while opening
  2. then call `flush_wal`

However, implementing that change doesn't give us the performance benefit, because most of the work is in creating the WAL in the first place, not flushing the changes to the OS page cache. I think we need to move `db.write` to the background thread altogether.

gabriele-0201:
I have done some empirical tests and:

with `diable_wal` during writes: https://share.firefox.dev/49QF9WY
with `set_manual_wal_flush` while opening: https://share.firefox.dev/3Jx4mL4

both use `nomt-bg-flusher` to perform `flush_wal` but the first one does not flush anything while the second one does flush, but during commit it spend anyway most of the time in `WriteToWal`

PR #153, opened 2024-04-23
opt: use rocksdb merge operator to write single nodes

rphmeier:
The idea here is to use RocksDB [merge operators](https://github.com/facebook/rocksdb/wiki/Merge-Operator)
to avoid writing full pages when they are only slightly updated. In large commits, we update lots of pages, but many of them by only a tiny amount. However, the pages at the lowest levels of the trie are usually almost completely altered, so diffing is wasteful.

Whenever we update fewer than `FULL_PAGE_THRESHOLD=32` nodes in a page, we simply write a diff to RocksDB to be handled by the merge operator. When we update at least that many nodes, we opt to write the whole page. This parameter will require some tuning, but 32 gave good results on my end.

This eliminates most of the overhead of the WAL on pages reduces the time spent in `rocksdb_write` by about half again after #157 on my machine.

### DIFF
--- a/nomt/src/store.rs
+++ b/nomt/src/store.rs
@@ -4,7 +4,8 @@ use nomt_core::{
     page_id::PageId,
     trie::{KeyPath, Node, TERMINATOR},
 };
-use rocksdb::{ColumnFamilyDescriptor, WriteBatch, DB};
+use rocksdb::{ColumnFamilyDescriptor, MergeOperands, WriteBatch, DB};
+use std::collections::HashMap;
 use std::sync::Arc;
 
 static FLAT_KV_CF: &str = "flat_kv";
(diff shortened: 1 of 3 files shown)


rphmeier:
* **#153** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/153?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

nomt/src/store.rs
```diff
@@ -130,3 +150,48 @@ impl Transaction {
...
+            let slot_end = slot_start + 32;
+            val[slot_start..slot_end].copy_from_slice(&op[1..]);
+        }
+    }
+
+    // merge operators are not supposed to fail unless there is corruption.
+    val
+}
+
+fn partial_merge_page_ops(
```
gabriele-0201, line 171: I would add a comment here explaining that this is merging multiple merge operations into a new one. It's not obvious at first glance that `prev_value` is an older operation and `operands` are possibly multiple new operations

gabriele-0201 (approved):
Great improvement!

nomt/src/store.rs
rphmeier: yes, good point. I will link to the the RocksDB documentation that makes this clear.

PR #154, opened 2024-04-24
bench: fix initialize backend only once

gabriele-0201:
#149 introduces a useful feature, but it comes with the side effect of causing all workloads to be executed on top of the previous one.

This pull request changes how workload initialization and execution processes run: initialization runs on a database, while the execution runs on a copy of the initialized database.

There is redundant code between the `open` and `open_copy` functions within each backend due to an issue with "`backend::Db` not being able to be turned into an object" when attempting to move `open` inside the trait.

Suggestions for avoiding the redundant code are welcome!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ profile.json
 /nomt/test
 
 # xtask
-/benchtop/sov_db
-/benchtop/nomt_db
-/benchtop/sp_trie_db
+/benchtop/sov_db*
+/benchtop/nomt_db*
+/benchtop/sp_trie_db*
 /benchtop/target
--- a/benchtop/src/backend.rs
+++ b/benchtop/src/backend.rs
@@ -21,6 +21,10 @@ pub enum Action {
 
 /// Trait implemented by all backends who wants to be benchmarked.
 pub trait Db {
+    /// Create a new backend using a copy of the database.
+    /// Delete any existing copies beforehand
+    fn open_copy(&self) -> Box<dyn Db>;
+
     /// Apply the given actions to the storage, committing them
     /// to the database at the end.
     ///
@@ -35,6 +39,14 @@ pub trait Db {
     /// Other spans can be measured by each backend, leaving space
     /// for more detailed tasks specific to each backend.
     fn apply_actions(&mut self, actions: Vec<Action>, timer: Option<&mut Timer>);
+
+    /// Apply the actions to a copy of the backend to measure performance without altering
+    /// the original database structure. This allows for applying and reverting actions
+    /// iteratively to measure performance.
+    fn apply_and_revert_actions(&self, actions: Vec<Action>, timer: Option<&mut Timer>) {
+        let mut revert_db = self.open_copy();
+        revert_db.apply_actions(actions, timer);
+    }
 }
 
 impl Backend {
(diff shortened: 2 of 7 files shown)


gabriele-0201:
* **#154** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/154?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

rphmeier (commented):
what's the purpose of making a copy?

why not perform the runs in the same DB?

The other thing that is missing IMO is that we should generate a fresh random workload every time rather than re-running the same workload each time.

gabriele-0201:
> what's the purpose of making a copy?

Otherwise, each execution of the workload would affect the underlying database, and the modifications would be seen in subsequent workload executions. Currently, starting from #149, if the initial additional size is 0, the first workload will write 20k balances. Subsequently, the second iteration will operate on a database with the combined initial capacity and writes from the previous workload. This results in iterations of workloads on different databases

> why not perform the runs in the same DB?

Cloning the entire folder makes it really easy to revert modifications for all currently supported databases

> The other thing that is missing IMO is that we should generate a fresh random workload every iteration rather than re-running the same workload each time.

Yes, this is completely doable if you are talking about the read/writes being performed, not the initial additional capacity. I can make it in a follow up

rphmeier:
> Otherwise, each execution of the workload would affect the underlying database, and the modifications would be seen in subsequent workload executions. Currently, starting from https://github.com/thrumdev/nomt/pull/149, if the initial additional size is 0, the first workload will write 20k balances

This seems OK, though, if we look at it as a fio style tool.

rphmeier:
And another thing on that point - fio allows for the user to specify a few different ways the benchmark runs:
  1. until interrupt
  2. until a certain amount of total data is written
  3. until a certain amount of time has passed

I think we should rework the workload system to achieve similar goals.
What we are calling a "workload" now is actually a transaction. It is often more interesting to see what happens when you perform similar transactions repeatedly for a certain amount of time or until a certain amount of total operations have been performed.

So, for example, we could say "run transactions of 1000 random RWs" until 30 seconds have passed or we have done total 100,000 RWs.

gabriele-0201:
> Benchmarking a single transaction vs. benchmarking the results of performing many transactions in sequence

Are you using "transaction" here to refer to what we describe as a "workload"?

Because currently, a "workload" is a set of actions that will be performed on the backend, where actions like read and write are considered transactions. In this context, a workload is already a sequence of many transactions. So maybe you're not referring to this?

Maybe we can provide a simple flag to specify whether we want to execute workloads each time on a freshly initialized backend or if each workload should be executed on the previous one, so that caching should be part of the final average time

gabriele-0201:
> Benchmarking a single transaction vs. benchmarking the results of performing many transactions in sequence

if here "transaction" refers to what we describe as a "workload" then maybe we can provide a flag to specify whether we want to execute workloads each time on a freshly initialized backend or if each workload should be executed on the previous one, so that caching should be part of the final average time.

> What we are calling a "workload" now is actually a transaction ..... we could say "run transactions of 1000 random RWs" until 30 seconds have passed or we have done total 100,000 RWs

To be fair, I find the word "transaction" here confusing (as it has too many meanings) but other than that, I think that what you described, mixed with what was proposed above, would be nice. If you want, you can benchmark a workload/transaction execution on its own, otherwise, you can specify some  other "limiters"

The interrupt feature is somewhat peculiar here because we support multiple backends, so the behavior is not clear. Should we perform measurements on one backend, record how much time has passed before the interrupt, and then perform the same workload within the same timeframe on the other backends?

I will start implementing features 2 and 3 for now

rphmeier:
> Should we perform measurements on one backend, record how much time has passed before the interrupt, and then perform the same workload within the same timeframe on the other backends?

Good question...I think this points to only allow a single backend run within the program. Users running multiple backends can just sequence the commands in their shell or script.

PR #155, opened 2024-04-24
proofs: create multiproof from vec of PathProof

gabriele-0201:


core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+
+        // The goal is to traverse the entire tree
+        // formed by all path proofs and only collect the siblings
+        // that cannot be reconstructed in the future
+        //
+        // The traversal does not occur on the siblings themselves,
+        // but on the ordered set of key_paths within PathProofs.
+        //
+        // For example, take two key_paths starting from the first bit.
+        // If two keys share the same bit at index i, they will have the same sibling at index i.
```
rphmeier, line 51: If two keys have the same prefix of 'i' bits, they will have the same sibling at index i.

As written, this says `010` and `110` both have the same sibling at index 2.

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+        // but if two key_paths have different bits at an index, the key_paths are split
+        // based on the value at that index.
+        //
+        // If all key_paths share a bit at an index, that sibling is required;
+        // if at least two key_paths differ, no sibling is needed, but the key_paths must be divided
+        // based on having bit 0 or 1 at that index.
+        //
+        // Iterate this algorithm on the bisection to determine the minimum necessary siblings.
+        //
+        // `external_siblings` whill follow this structure for each bicection:
```
rphmeier, line 68: typos: "whill" -> "will", "bicection" -> "bisection"

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+    // All subpaths related to a single terminal node
+    sub_paths: Vec<SubPathProof>,
+    // Vector containing the minimum number of nodes required
+    // to reconstruct all other nodes later.
+    external_siblings: Vec<Node>,
+}
+
+impl MultiProof {
+    /// Construct a MultiProof from a vector of *ordered* PathProof
+    pub fn from_path_proofs(path_proofs: Vec<PathProof>) -> Self {
```
rphmeier, line 35: this function would be more readable if it were split into multiple helper functions.

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+                sub_paths.push(sub_path_proof);
+
+                // insert all collected siblings in this sub tree in ext_sibling
+                // at the position this bisection is pointing at
+                for sibling in ext_sibling_in_sub_tree.into_iter().rev() {
+                    external_siblings.insert(curr_traverse_info.ext_sibling_index, sibling);
+                }
+                ext_sibling_in_sub_tree = vec![];
+
+                // skip to the next bisection in the stak, if empty we're finished
```
rphmeier, line 121: stack

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+        };
+
+        loop {
+            // Based on how we bisect the key_path slice, we can see that each TraverseInfo
+            // will represent a slice where all key_paths will have same bits before
+            // TraverseInfo.path_index.
+
+            // If there is no longer a common prefix, there is only one key in the TraverseInfo,
+            // and then all remaining siblings are required for the multiproof.
+            if curr_traverse_info.path_slice_lower == curr_traverse_info.path_slice_upper {
```
rphmeier, line 99: this whole block could be extracted to a function like `fn prove_unique_path_remainder`

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+        //
+        // `external_siblings` whill follow this structure for each bicection:
+        // |commons siblings| ext siblings in the right bisection | ext siblings in the left bisection |
+
+        // it represent a slice of the vector of key_paths with the notion of
+        // path_index, at which bit in the key_path that slice is pointing at
+        struct TraverseInfo {
+            path_index: usize,
+            path_slice_lower: usize,
+            path_slice_upper: usize,
```
rphmeier, line 76: it seems like this is an inclusive bound?

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+        // based on the value at that index.
+        //
+        // If all key_paths share a bit at an index, that sibling is required;
+        // if at least two key_paths differ, no sibling is needed, but the key_paths must be divided
+        // based on having bit 0 or 1 at that index.
+        //
+        // Iterate this algorithm on the bisection to determine the minimum necessary siblings.
+        //
+        // `external_siblings` whill follow this structure for each bicection:
+        // |commons siblings| ext siblings in the right bisection | ext siblings in the left bisection |
```
rphmeier, line 69: typo: common

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+        // because it can be reconstructed from siblings collected later in the other key.
+        // When a differing bit is found, each key_path needs separate scanning.
+        //
+        // When there is a single key_path, all siblings are necessary.
+        //
+        // When there are more than two key_paths, the same logic is applied,
+        // but if two key_paths have different bits at an index, the key_paths are split
+        // based on the value at that index.
+        //
+        // If all key_paths share a bit at an index, that sibling is required;
```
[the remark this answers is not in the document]
rphmeier: or is this 'common siblings'?
or both, depending on how you look at it?

probably a good idea to clean up the terminology.

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+                        .collect(),
+                };
+                sub_paths.push(sub_path_proof);
+
+                // insert all collected siblings in this sub tree in ext_sibling
+                // at the position this bisection is pointing at
+                for sibling in ext_sibling_in_sub_tree.into_iter().rev() {
+                    external_siblings.insert(curr_traverse_info.ext_sibling_index, sibling);
+                }
+                ext_sibling_in_sub_tree = vec![];
```
rphmeier, line 119: you can avoid this with `for sibling in ext_sibling.drain(..).rev()` in the line above

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+            ext_sibling_index: usize,
+        }
+        let mut stack_traverse_info: Vec<TraverseInfo> = vec![];
+
+        // External siblings encountered during a subtree traversal
+        let mut ext_sibling_in_sub_tree: Vec<Node> = vec![];
+
+        // initially we're looking at all the key_paths
+        let mut curr_traverse_info = TraverseInfo {
+            path_index: 0,
```
rphmeier, line 86: I find the name `path_index` misleading. It looks like it'd be the index in `path_proofs`, but it's actually the bit index in the key path where the bisection was performed.
I recommend changing this to `bit_index` or `path_bit_index`.

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+                .key_path()
+                .view_bits::<Msb0>();
+            let path_upper = path_proofs[curr_traverse_info.path_slice_upper]
+                .terminal
+                .key_path()
+                .view_bits::<Msb0>();
+
+            if path_lower[curr_traverse_info.path_index]
+                != path_upper[curr_traverse_info.path_index]
+            {
```
rphmeier, line 145: and this whole block could be extracted to a function like "begin_bisect"

rphmeier (commented):
Okay, finally reviewed this.

I like the algorithm - it seems clean, though I haven't looked at how it's verified yet. I expect that the recursive format `[common siblings | right branch siblings | left branch siblings]` will be complicated to handle, but I may be wrong, and it could be easily addressable with a recursive verification algorithm.

I also would prefer not to see re-implementation of algorithms like binary search in our code unless absolutely necessary (for performance or limitations of the std APIs). There is no driving performance need in multi-proof creation (yet). If anything, a performance bottleneck will be verification, which may run in compute-constrained environments like Risc0. So keep in mind to optimize code for readability wherever possible!

The "one giant function" is pretty horrible, though, so I really strongly recommend at least extracting some of the functionality out into standalone functions, or even better, into a struct that encapsulates the shared state. You can just have `MultiProof::from_path_proofs` use this struct internally.

Looking at the algorithm, because it's depth-first it doesn't seem strictly necessary to have all the paths before beginning the multi-proof. So, looking forward, there may be a way to consume one terminal at a time while doing the update logic.

core/src/proof.rs
```diff
@@ -6,12 +6,30 @@ use crate::trie::{
 
 use bitvec::prelude::*;
 
+/// Wrapper for a terminal node, it will store the LeafData if it is a leaf node,
+/// and just the KeyPath to that terminal if it is a terminator node
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PathProofTerminal {
+    Leaf(LeafData),
+    Terminator(KeyPath),
```
rphmeier, line 14: It doesn't make sense for a Terminator to have a KeyPath. It can have a bit-path, but not a full key-path, unless it's at depth 256.

core/src/proof.rs
```diff
@@ -6,12 +6,30 @@ use crate::trie::{
...
+}
+
+impl PathProofTerminal {
+    /// Return the KeyPath of the Terminal Node
+    pub fn key_path(&self) -> &[u8; 32] {
```
rphmeier, line 19: likewise, with above, this can return a bit-slice but not a key-path, because terminators don't have full key-paths.

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+
+// Each terminal node may have a set of siblings that are uniquely
+// used during the verification with its terminal node.
+// This struct contains the terminal node itself,
+// the depth at which the siblings start to be uniquely associated to that terminal,
+// and the siblings themselves.
+#[derive(Debug, Clone)]
+struct SubPathProof {
+    terminal: PathProofTerminal,
+    depth: u8,
```
rphmeier, line 19: does this suffer from the same problem as trie_pos, where we might want depth = 256 and it could be out of range for u8?

maybe better to use a usize

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+                        .iter()
+                        .skip(curr_traverse_info.path_index)
+                        .copied()
+                        .collect(),
+                };
+                sub_paths.push(sub_path_proof);
+
+                // insert all collected siblings in this sub tree in ext_sibling
+                // at the position this bisection is pointing at
+                for sibling in ext_sibling_in_sub_tree.into_iter().rev() {
```
rphmeier, line 116: `ext_sibling_in_sub_tree` is guaranteed empty here because terminal nodes always follow immediately after internal nodes. you can omit this code.

core/src/multi_proof.rs
gabriele-0201: oh yes, that's wrong, it should probably be:

If two keys share bits up to index i, they will have the same sibling at index i.

core/src/multi_proof.rs
gabriele-0201: Yes, a comment would definitely help. I will add it

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+                // binary search between key_paths in the slice to see where to
+                // perform the bisection
+
+                // path_slice_upper and path_slice_lower will never be the same otherwise
+                // bits at path_index would have been the same
+
+                let mut low = curr_traverse_info.path_slice_lower;
+                let mut up = curr_traverse_info.path_slice_upper;
+                let mut mid;
+                loop {
```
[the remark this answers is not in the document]
gabriele-0201: Yes, you're totally right. That's an example of premature optimization, where I thought that since the vector was sorted, I could have done it better than the compiler.

Using `slice::binary_search` or one of its variations is not doable here because the first 1 is required.

As you said, a solution like this seems to be the cleanest and most reliable one (and it could be improved later on)

```rust
                let mid = curr_traverse_info.lower
                    + path_proofs[curr_traverse_info.lower..=curr_traverse_info.upper]
                        .iter()
                        .position(|path_proof| {
                            path_proof.terminal.key_path().view_bits::<Msb0>()
                                [curr_traverse_info.path_bit_index]
                        })
                        .expect("There must be at least one 1 bit");

```

core/src/multi_proof.rs
gabriele-0201: Ok, you're totally right in saying that `ext_sibling_in_sub_tree` is guaranteed to be empty here; however, I didn't understand your explanation.

What I'm thinking is that if we have found a unique sub-path, then we can be sure that there can't be `ext_sibling_in_sub_tree` because they were already emptied in the last bisection that brought us to this unique sub-path

core/src/multi_proof.rs
gabriele-0201: This has been removed, but the same trick is applied in another section of the code!

core/src/multi_proof.rs
```diff
@@ -0,0 +1,548 @@
...
+                        } else {
+                            up = mid - 1;
+                        }
+                    }
+                }
+
+                // insert all collected siblings (those are the shared ones)
+                // in this sub tree in ext_sibling at the position this bisection is pointing at
+                let collected_siblings = ext_sibling_in_sub_tree.len();
+                for sibling in ext_sibling_in_sub_tree.into_iter().rev() {
```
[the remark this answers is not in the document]
gabriele-0201: nice! thanks!

core/src/multi_proof.rs
gabriele-0201: yes, you're totally right. There could be a case where depth is 256, for example if two keys with only the lsb as different bit then there would be two `TraverseInfo` with `path_bit_index` equal to 256 but it would not be used as index for the key_path because by construction those `TraverseInfo` would represent a single `path_proof` and not a range.  

I have added also another test to cover this scenario `test_multiproof_creation_two_path_proofs_256_depth`

core/src/proof.rs
```diff
@@ -6,12 +6,30 @@ use crate::trie::{
...
+/// and just the KeyPath to that terminal if it is a terminator node
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PathProofTerminal {
+    Leaf(LeafData),
+    Terminator(BitVec<u8, Msb0>),
+}
+
+impl PathProofTerminal {
+    /// Return the KeyPath of the Terminal Node
+    pub fn key_path(&self) -> &BitSlice<u8, Msb0> {
```
rphmeier, line 19: nit: for a follow-up, I'd prefer if this were named `path` rather than `key_path`. we haven't actually defined this convention anywhere, but I treat "key path" as a full 32-byte path (based off of a _key_) and a "path" as the bit-path to any node in the trie.

core/src/multi_proof.rs
```diff
@@ -0,0 +1,581 @@
...
+                let mid = curr_traverse_info.lower
+                    + path_proofs[curr_traverse_info.lower..=curr_traverse_info.upper]
+                        .binary_search_by(|path_proof| {
+                            if !path_proof.terminal.key_path()[curr_traverse_info.path_bit_index] {
+                                std::cmp::Ordering::Less
+                            } else {
+                                std::cmp::Ordering::Greater
+                            }
+                        })
+                        .expect_err("There must be at least one bit set to 1");
```
rphmeier, line 162: for a follow-up: codebase style is to prefer `unwrap` with a comment.

rphmeier (approved):

core/src/proof.rs
gabriele-0201: this makes perfectly sense! thanks!

core/src/multi_proof.rs
gabriele-0201: oh, then I fell behind. Didn't we discuss adding a brief description in `expect` plus a comment explaining why this is impossible? 

In any case, if you made any new decisions, I will change it in a follow-up 👍🏻

PR #161, opened 2024-04-26
feat: introduce a RegionedWritePass which allows write-locking specific regions of data

rphmeier:
This introduces a `RegionedWritePass` that can be split according to an implementation of the
`Region` trait. I don't really care about the abstraction here, but I followed the lead of
`RwPassCell` in being generic and made a generic implementation.

I have a very particular concrete use-case in mind for this: parallel trie update.

The idea is that we have N worker threads, and we give them each 1/N of the read/writes to process.
We can use the `RegionedWritePass` as not only protection for writes, but also as a way to easily
schedule the update. I'll elaborate.

We give each worker thread a pass corresponding to the sections of the trie they are responsible. The domain is a `min, max: PageId`. This pass has a parent which corresponds to the next point of contention with another worker thread, and so on, all the way up to the root page. Since only the last worker to `consume` their write pass gets the parent
pass, this gives us an easy scheduling protocol for free: whichever worker finished their part of the work last gets to continue.

### DIFF
--- a/nomt/src/rw_pass_cell.rs
+++ b/nomt/src/rw_pass_cell.rs
@@ -25,8 +25,12 @@
 
 use std::{
     cell::UnsafeCell,
+    marker::PhantomData,
     ops::{Deref, DerefMut},
-    sync::{Arc, Weak},
+    sync::{
+        atomic::{AtomicBool, AtomicUsize, Ordering},
+        Arc, Weak,
+    },
 };
 
 use parking_lot::{RawRwLock, RwLock};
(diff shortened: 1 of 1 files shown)


rphmeier:
* **#161** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/161?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

gabriele-0201 (commented):
Wow, super nice! I understood how `RegionedWritePass` is created and can be split into two separate regions, and how `remaining_count` is used to implicitly create synchronization primitives between different regions being written.

I couldn't understand how and why `sync_dummy` is the key to handling syncronization between threads. This is likely due to my lack of knowledge in Atomic Ordering, and probably also the `Send` and `Sync` traits!

It seems super promising as a primitive to handle parallel trie update!

rphmeier:
Thanks! Yes, I really hope to make good use of it in the parallel trie update.

I want to rework the atomics slightly before merging, but the core idea is this:
  1. When you hand in a regioned write pass and get the parent, you need to acquire a lock on all of the pages covered by the parent
  2. The writes done by sibling `RegionedWritePasses` need to be made available on your thread as a result.
  3. The thread which sends the pass, either by being dropped, handed in, or sent through an envelope needs to write a value with Ordering::Release, and the thread which receives the pass through a hand-in or envelope needs to read that corresponding value with Ordering::Acquire.

gabriele-0201 (approved):
Thank you for the explanations! After doing some reading on how ordering works, I now understand the synchronization mechanism you implemented!

nomt/src/rw_pass_cell.rs
```diff
@@ -237,3 +301,190 @@ impl<'a, 'pass, T> DerefMut for WriteGuard<'a, 'pass, T> {
...
+        } else {
+            None
+        }
+    }
+}
+
+impl<R> Drop for RegionedWritePass<R> {
+    fn drop(&mut self) {
+        if let Some(ref parent) = self.parent {
+            // SAFETY: release our writes to other threads.
```
pepyakin, line 486: safety notes typically pertain to an unsafe block which there are none.

pepyakin (approved):

nomt/src/rw_pass_cell.rs
rphmeier: it has critical impacts on the safety of other unsafe blocks, though. omitting this code would cause unsafety later

PR #166, opened 2024-04-29
implement a `PageRegion` structure

rphmeier:
This implements a `PageRegion` that fulfills the requirements of the `rw_pass_cell::Region` trait.

### DIFF
--- a/nomt/src/page_region.rs
+++ b/nomt/src/page_region.rs
@@ -1,0 +1,275 @@
+//! Implements a regioning scheme for the page tree for use as a [`crate::rw_pass_cell::Region`].
+//! This enables safe, sharded, mutable access to portions of the page tree.
+//!
+//! Each [`PageRegion`] spans an entire sub-section of the page tree for exclusive access.
+//! This is either a page P and all of its possible descendants, or an inclusive range of a
+//! page P's child pages and all of their descendants.
+//!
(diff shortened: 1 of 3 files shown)


nomt/src/page_region.rs
```diff
@@ -0,0 +1,259 @@
...
+use crate::rw_pass_cell::{Region, RegionContains};
+
+/// A region of pages for shared/exclusive access.
+///
+/// See module docs for more details.
+#[derive(Debug, Clone)]
+pub struct PageRegion {
+    // non-exclusive access: if min is `None` any page along the path to this one (excluding this).
+    //                       otherwise, including this page.
+    // exclusive access: if min is `None`, just to this page ID. otherwise none.
```
gabriele-0201, line 52: shouldn't it be: `exclusive access: if min is None, from this page ID. otherwise from min`

gabriele-0201 (approved):

gabriele-0201:
The exclusive_min option is really nice for indicating whether the pointed page has exclusive access or not! Not talking about PageId ordering, it makes the implementation super clean!

nomt/src/page_region.rs
rphmeier: ah, this comment was meant to just be on this variable, but I will edit for clarity

pepyakin (approved):
Ideally we settle on the same terminology re shared/non-exclusive & exclusive and read & write/mutable.

PR #172, opened 2024-05-06
refactor: extract seek to its own module

rphmeier:


pepyakin (commented):
I think this needs a rebase/restack

PR #177, opened 2024-05-12
rw_pass_cell: add custom drop to write_pass

gabriele-0201:


gabriele-0201:
* **#177** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/177?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

nomt/src/rw_pass_cell/mod.rs
```diff
@@ -256,6 +263,17 @@ impl<R: Region + Clone> WritePass<R> {
     }
 }
 
+impl<R> Drop for WritePass<R> {
+    fn drop(&mut self) {
+        if let Some(ref parent) = self.parent {
+            if !self.consumed {
+                // SAFETY: release our writes to other threads.
+                parent.remaining_children.fetch_sub(1, Ordering::AcqRel);
```
rphmeier, line 271: you only need `Release` here

gabriele-0201: Interesting, it works, but I don't understand why `Release` is enough. If `load-acquire` combined with `store-release` synchronizes the accessed and modified data, how is it that in `drop` we only need `store-release` with a `load-relaxed` (that's the behavior of `fetch_sub` with only `Ordering::Release`). How can the `load-relaxed` be reliable?

I have a possible idea but I don't know if it makes sense or not. Is it because there are only two write passes that share the same `remaining_children` atomic, and thus:

If the first one is consumed, then we don't really care about the value being loaded because it will be dropped. However, if it is first dropped, then we can load the value with relaxed (because nothing is happening simultaneously) and write it with `Release` to define an "order" that will be respected by the `AcqRel` in consume

rphmeier: Remember that the operation is still atomic, and the ordering relates to propagation of store operations performed on the same thread.

Load-relaxed is always reliable _as it relates to the single atomic variable_. We only use the orderings to make sure that our writes to the `UnsafeCell` propagate to other threads. In `Drop`, we will make no further writes, so we don't need to `Acquire` any writes from other threads. We just need to `Release` to any thread that might continue writing to those cells.

PR #184, opened 2024-05-13
refactor: benchtop workloads unique each iteration; remove unused code

rphmeier:
I reworked benchtop to produce a new workload each iteration rather than applying the same actions
each time.

This actually brings the sp-trie benchmark more in line with NOMT on commit, because it has a fast
path for when trie nodes aren't altered, and this is what we were seeing before.

I think this will lead to much more realistic testing.

I also removed a few things:
  1. Sequential workload - because we hash keys, they're not sequential. I still would like to have
     this at some point, but it doesn't make sense until we don't hash keys.
  2. Action - because generating fresh actions each time is really what we should be doing.

### DIFF
--- a/benchtop/src/sp_trie.rs
+++ b/benchtop/src/sp_trie.rs
@@ -1,13 +1,10 @@
-use crate::{
-    backend::{Action, Db},
-    timer::Timer,
-};
+use crate::{backend::Transaction, timer::Timer, workload::Workload};
 use hash_db::{AsHashDB, HashDB, Hasher as _, Prefix};
 use kvdb::KeyValueDB;
 use kvdb_rocksdb::{Database, DatabaseConfig};
 use sha2::Digest;
 use sp_trie::trie_types::TrieDBMutBuilderV1;
-use sp_trie::DBValue;
+use sp_trie::{DBValue, LayoutV1, TrieDBMut};
 use std::collections::HashMap;
 use std::sync::Arc;
 use trie_db::TrieMut;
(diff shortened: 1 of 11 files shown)


rphmeier:
* **#184** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/184?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

benchtop/src/sp_trie.rs
```diff
@@ -106,39 +80,27 @@ impl Db for SpTrieDB {
...
-                    let key_path = sha2::Sha256::digest(key);
-
-                    let _timer_guard_read = timer.as_mut().map(|t| t.record_span("read"));
-                    trie_db_mut
-                        .get(&key_path)
-                        .expect("Impossible fetching from sp-trie db");
-                }
-            }
-        }
+        let _timer_guard_commit = {
```
gabriele-0201, line 83: why this scope?

gabriele-0201 (approved):
nice refactor, I like how the `Transaction` and `Workload` traits interact!

benchtop/src/sp_trie.rs
rphmeier: the scope is required for the borrow of `new_root` to end. moving the new_root borrow into the `Tx` struct confuses Rust's Non-Lexical Lifetime checker so we have to fall back to old-school lexical lifetimes :)

PR #193, opened 2024-05-16
add optional metrics collector

gabriele-0201:
measure page_requests_count, page_cache_misses, page_fetch_time, and value_fetch_time

The decision I have made for now is to collect metrics based on nomt instance. Metrics are collected across all sessions and printed on-demand. If you prefer to collect them per session, it would be easy to update. Thinking about #195, it would probably be nice to remove the initialization from the metric collection. This can be addressed in a follow-up

example outcome:
```shell
     Running `target/debug/benchtop bench sequential -b nomt -n randrw -s 25000 -c 15 --op-limit 100000`
metrics:
  page requests         747026
  page cache misses     174420 - 23.35% of page requests
  page fetch mean       10232 ns
  value fetch mean      8437 ns
nomt
  mean workload: 2535 ms
  mean read: 9712 ns
  mean commit_and_prove: 1946 ms
```

### DIFF
--- a/benchtop/src/backend.rs
+++ b/benchtop/src/backend.rs
@@ -72,4 +72,11 @@ impl DB {
             }
         }
     }
+
+    pub fn print_metrics(&self) {
+        match self {
+            DB::Nomt(db) => db.print_metrics(),
+            _ => (),
+        }
+    }
 }
--- a/nomt/src/metrics.rs
+++ b/nomt/src/metrics.rs
@@ -1,0 +1,131 @@
+use std::sync::{
+    atomic::{AtomicU64, Ordering},
+    Arc,
+};
+
+/// Metrics collector, if active, it provides Counters and Timers
+#[derive(Clone)]
+pub enum Metrics {
+    Active(Arc<ActiveMetrics>),
+    Inactive,
+}
+
+impl Metrics {
+    /// Returns the Metrics object, active or not based on the specified input
+    pub fn new(metrics: bool) -> Self {
+        if metrics {
+            Metrics::Active(Arc::new(ActiveMetrics {
+                page_requests_count: AtomicU64::new(0),
+                page_cache_misses_count: AtomicU64::new(0),
+                page_fetch_time: Timer::new(),
+                value_fetch_time: Timer::new(),
+            }))
+        } else {
+            Metrics::Inactive
+        }
+    }
+
+    /// Print collected metrics to stdout
+    pub fn print(&self) {
+        match self {
+            Metrics::Active(metrics) => {
+                println!("metrics");
+
+                let tot_page_requests = metrics.page_requests_count.load(Ordering::Relaxed);
+                println!("  page requests         {}", tot_page_requests);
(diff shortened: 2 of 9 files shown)


gabriele-0201:
* **#193** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/193?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

nomt/src/metrics.rs
```diff
@@ -0,0 +1,152 @@
...
+            fn drop(&mut self) {
+                let elapsed = self.start.elapsed().as_nanos() as u64;
+                *self.measure = elapsed;
+            }
+        }
+
+        // Safety: at most multiple threads may read from the same index
+        // resulting in potential records loss when writing to that index concurrently
+        //
+        // TODO: Is this really thread-safe?
```
rphmeier, line 138: yeah, this is not thread-safe.

nomt/src/metrics.rs
```diff
@@ -0,0 +1,131 @@
...
+            Metrics::Inactive => {
+                println!("Metrics collection was not activated")
+            }
+        }
+    }
+}
+
+/// Active metrics that can be collected during execution.
+pub struct ActiveMetrics {
+    pub page_requests_count: AtomicU64,
```
rphmeier, line 67: for a follow-up: docs on public items

benchtop/src/backend.rs
```diff
@@ -49,4 +49,11 @@ impl DB {
             DB::Nomt(db) => db.execute(timer, workload),
         }
     }
+
+    pub fn print_metrics(&self) {
```
rphmeier, line 53: for a follow-up: docs on public items.

I would also prefer to have an approach like:

```
impl Nomt {
  fn metrics(&self) -> Metrics { ... }
}

// user code
nomt.metrics().print();

// or
send_to_prometheus(nomt.metrics())

// or whatever
```

rphmeier (approved):

nomt/src/page_cache.rs
```diff
@@ -435,6 +439,12 @@ impl PageCache {
     /// If the page is already in the cache, this method does nothing. Otherwise, it fetches the
     /// page from the underlying store and caches it.
     pub fn prepopulate(&self, page_id: PageId) {
+        if let Metrics::Active(metrics) = &self.shared.metrics {
```
rphmeier, line 442: for a follow-up: I think these `match`es should be encapsulated inside the metrics type itself, with methods for bumping counts, creating timer guards, etc.

gabriele-0201: Yes, I tried to pursue this, but I couldn't do it easily. That's because currently, from that match, we're able to directly access the Atomic types. If we wrap them inside any other type, such as a HashMap to simplify the access to metrics, we end up with a Send/Sync problem.

I have an idea of using enum + some glue code inside of metrics! I will push it soon!

PR #195, opened 2024-05-16
refactor: remove most commands from benchtop

rphmeier:
We had a bunch of commands like "run", "init", "bench isolate", etc. that we were never using
in practice.

I have
  1. altered workloads so they can run in multiple iterations (for example, to instantiate large DBs)
  2. moved everything logically into the Run command, which now can take a `--no-reset` flag to use the existing DB


rphmeier:
* **#195** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/195?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

benchtop/src/backend.rs
```diff
@@ -41,12 +41,35 @@ pub enum DB {
...
-    /// Execute some code against the DB using the given closure.
-    pub fn execute(&mut self, timer: Option<&mut Timer>, workload: &mut dyn Workload) {
-        match self {
-            DB::Sov(db) => db.execute(timer, workload),
-            DB::SpTrie(db) => db.execute(timer, workload),
-            DB::Nomt(db) => db.execute(timer, workload),
+    /// Execute a workload repeatedly until done.
+    pub fn execute(&mut self, mut timer: Option<&mut Timer>, workload: &mut dyn Workload) {
+        while !workload.is_done() {
+            let timer = timer.as_mut().map(|x| &mut **x);
```
gabriele-0201, line 47: here you could consider using `let timer = timer.as_deref_mut();`

benchtop/src/main.rs
```diff
@@ -1,55 +1,77 @@
...
+        params.limits.ops.unwrap_or(u64::max_value()),
     )?;
 
-    let mut db = Backend::Nomt.instantiate(true, params.fetch_concurrency);
-    db.execute(None, &mut *workload);
+    let mut db = params
+        .backend
+        .instantiate(!params.no_reset, workload_params.fetch_concurrency);
+    if !params.no_reset {
+        db.execute(None, &mut *init);
```
gabriele-0201, line 63: It's not obvious that if `no_reset = false`, then the workload initialization would be applied; otherwise, it wouldn't. It's probably worth mentioning this in the cli

benchtop/src/cli.rs
```diff
@@ -63,21 +33,48 @@ impl Display for Backend {
...
+    pub limits: RunLimits,
+
+    /// The backend to run the workload against.
+    #[arg(required = true, long, short)]
+    pub backend: Backend,
+
+    /// Whether to reset the database.
+    #[clap(default_value = "true")]
+    #[arg(long, short)]
+    pub no_reset: bool,
```
gabriele-0201, line 63: Wouldn't it be easier to have a `reset` flag defaulting to false? 
This way, logically adding the flag `--reset` would trigger a database reset

gabriele-0201 (approved):
wow, 90% of benchtop functionalities are gone, but if those are the minimum requirements, I'm ok to make it simpler!

benchtop/src/cli.rs
rphmeier: yeah, true.

benchtop/src/backend.rs
rphmeier: nice tip

PR #206, opened 2024-05-19
refactor: metrics are referenced with enums

gabriele-0201:
This PR aims to address the following comment: https://github.com/thrumdev/nomt/pull/193#discussion_r1605440239

The main problem is handling interior mutability while working with Sync types. ~~This solution utilizes immutable structs so that metrics can be accessed from multiple threads and allows working with atomic types.~~

The names `Metrics` and `Metric` may not be the best choices, but I couldn't think of alternatives. Suggestions are welcome!

### DIFF
--- a/nomt/src/lib.rs
+++ b/nomt/src/lib.rs
@@ -3,7 +3,7 @@
 //! A Nearly-Optimal Merkle Trie Database.
 
 use bitvec::prelude::*;
-use metrics::Metrics;
+use metrics::{Metric, Metrics};
 use std::{
     mem,
     rc::Rc,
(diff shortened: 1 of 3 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/206?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#206** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/206?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#205** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/205?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

rphmeier:
A HashMap is not the right solution here. HashMap is for key-value pairs which are determined at runtime, or otherwise large.

We have a handful of metrics we want to collect - they just should be fields as before.

gabriele-0201:
Now, `ActiveMetrics` does not use an `HashMap` anymore, but metrics are just fields. However, it uses an enum `Metric` to specify the inner field from the public struct `Metrics`. If using an enum to implement the required abstractions is not okay, then I don't know yet which way to proceed

rphmeier (approved):

rphmeier:
Much better now!

> If using an enum to implement the required abstractions is not okay, then I don't know yet which way to proceed

One approach could've been to add multiple functions `record_one_metric` and `record_another_metric`. But this is also fine.

PR #215, opened 2024-05-23
nomt: add commit without witness collection

### LINKED ISSUE #210: non-proving mode
Not every node requires collecting witnesses. Consider adding a configuration option to disable this behavior, possibly adjusting the API.

gabriele-0201:
I envisioned two main possibilities to add the option of *not collecting a witness during commit, as described in #210. One of them could be to add a field to the options and change the `commit_and_prove` method to just `commit` and make it return an enum, with or without a witness. 

Another possibility, the one I implemented, is to avoid specifying this in the options (as it is not required) but simply add a new method to the `nomt` API, `commit`, which will return only the new root after committing everything to the database.

Tests are still missing. If the approach is okay, I will address them in a follow-up

close #210

### DIFF
--- a/nomt/src/commit/mod.rs
+++ b/nomt/src/commit/mod.rs
@@ -148,9 +148,15 @@ impl Committer {
     }
 
     /// Commit the given key-value read/write operations. Key-paths should be in sorted order
-    /// and should appear at most once within the vector.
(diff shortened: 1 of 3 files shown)


gabriele-0201:
* **#215** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/215?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

rphmeier (approved):

PR #245, opened 2024-08-12
chore: move io module to NOMT main

rphmeier:
The main point of this PR is that we shouldn't have two copies of the io module, one in NOMT and one in bitbox. I chose to use the main NOMT location, but maybe having it in bitbox makes more sense. Something for a follow-up.


rphmeier:
### Merge activity

* **Aug 12, 4:06 PM EDT**: @rphmeier started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/245).

PR #266, opened 2024-08-13
fix: stop writeout from deadlocking when fsync submission fails

rphmeier:
In the BBN / LN writeout helpers, we would only attempt to submit an fsync once - at exactly the point we got the last page completion. If the queue was full, it would never be submitted again and therefore could never finish.

I observed this behavior often in the BBN routine but never in the leaf routine. I believe this is because BBN would attempt to submit the fsync while leaves were being written and clogging the io_uring worker channel. 

The fix is to adopt a `should_fsync` approach akin to what is done in the Meta helper.

### DIFF
--- a/nomt/src/beatree/writeout.rs
+++ b/nomt/src/beatree/writeout.rs
@@ -35,12 +35,14 @@ pub fn run(
                 remaining: bbn.len() + bbn_free_list_pages.len(),
                 bbn,
                 free_list_pages: bbn_free_list_pages,
+                should_fsync: false,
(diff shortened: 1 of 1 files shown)


rphmeier:
* **#266** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/266?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

gabriele-0201 (approved):

PR #269, opened 2024-08-14
beatree: change writeout workflow

gabriele-0201:
This is not the final stage for the writeout module; it functions in tests, but this pull request should be followed by either #270 or a variation of the `MetaSwap` struct

PR #270, opened 2024-08-14
bitbox: simplify writeout

gabriele-0201:
I thought that `IoDmux` was a bit over-complicated because we don't really need to worry about the order of writes during writeout; we just need them to occur.

An alternative to this pr is to modify `IoDmux` and use the same approach used in `LnWriteOut` and `BbnWriteOut` for `MetaSwap`

### DIFF
--- a/nomt/src/beatree/writeout.rs
+++ b/nomt/src/beatree/writeout.rs
@@ -26,7 +26,6 @@ pub fn run(
     ln_extend_file_sz: Option<u64>,
     new_meta: Meta,
 ) {
-    let now = std::time::Instant::now();
     let io = IoDmux::new(io_sender, io_handle_index, io_receiver);
     do_run(
         Cx {
@@ -43,15 +42,13 @@ pub fn run(
                 ln_remaining: ln.len(),
                 ln,
             },
-            meta_swap: Some(MetaSwap {
+            meta_swap: MetaSwap {
                 meta_fd,
                 new_meta: Some(new_meta),
-                should_fsync: false,
-            }),
+            },
         },
         io,
     );
-    println!("writeout {}us", now.elapsed().as_micros());
 }
 
 fn do_run(mut cx: Cx, mut io: IoDmux) {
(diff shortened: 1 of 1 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/270?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#270** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/270?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#269** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/269?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

nomt/src/beatree/writeout.rs
```diff
@@ -74,34 +71,25 @@ fn do_run(mut cx: Cx, mut io: IoDmux) {
...
-        // At this point, the BBN and LN files are fully synced. We can now update the meta file.
-        if let Some(ref mut meta_swap) = &mut cx.meta_swap {
-            if meta_swap.run(&mut io) {
-                cx.meta_swap = None;
-                return;
-            }
-        }
-    }
+    cx.bbn_write_out.wait_fsync(&mut io);
+    cx.ln_write_out.wait_fsync(&mut io);
```
rphmeier, line 78: seems better to dispatch the fsyncs at the same time and then handle the completions async.

for a follow-up.

rphmeier (changes requested):
I'm a bit conflicted around this PR.

On one hand, it does simplify things and looks like it is more aggressive with pushing onto io_uring. On the other, it looks like actually handling any kind of io error is going to be nearly impossible and would require either adding something like the inboxes back or to encapsulate the logic within `fn recv`. The inbox approach was better for that. Likewise, any attempt to handle completions _while_ write submission is ongoing would require another refactor of this code.

User data on io requests/completions is useful!

Even though it's simpler in the sense that there's less code, it's more conceptually complicated. 

The only refactor that was needed IMO is to move completion polling out of the `run` functions and have each helper gain a `handle_completion` function. And to have the main `run` loop attempt to take a single completion on each iteration, so long as each `run` indicated it wasn't currently trying to submit anything / the completion queue was full.

pepyakin:
Agreed. Closing for now just to clean up the queue, but feel free to resurrect as soon as you come back @gabriele-0201.

PR #298, opened 2024-08-20
io: retry read/write until complete

### LINKED ISSUE #288: io: retry read/write until complete
This issue exists both on our io_uring backend and when using `pread`/`pwrite` directly: we assume that a single `pread`/`pwrite` syscall always reads or writes the full page.

Though it's unlikely to occur, for correctness we need to retry requests whenever they don't fully read/write. Failing to do so could lead to some really nasty DB corruptions. In io_uring we should have something like a `retries: VecDeque<IoCommand>` that we prefer to load from before reading other incoming requests and for raw syscalls we should just call them in a loop.

I don't think we need a bunch of complex mechanics to only try to read/write the parts that were incomplete, because I don't expect incomplete I/O to be common enough to affect performance even if we completely re-do reads and writes.

gabriele-0201:
If a read or write operation is not fully completed, the command is reapplied

Errors should now be properly handled. However, this has revealed a problem with the fsync command. Each fsync call performed through io_uring is failing with an "Invalid Argument" error message.

A fix will be included in a follow-up pr

closes #288

### DIFF
--- a/nomt/src/io/linux.rs
+++ b/nomt/src/io/linux.rs
@@ -1,8 +1,11 @@
-use super::{CompleteIo, IoCommand, IoKind, IoPacket, PAGE_SIZE};
+use super::{CompleteIo, IoCommand, IoKind, IoKindResult, IoPacket, PAGE_SIZE};
 use crossbeam_channel::{Receiver, Sender, TryRecvError};
 use io_uring::{cqueue, opcode, squeue, types, IoUring};
 use slab::Slab;
-use std::time::{Duration, Instant};
+use std::{
+    collections::VecDeque,
+    time::{Duration, Instant},
+};
 
 const RING_CAPACITY: u32 = 128;
 
@@ -65,6 +68,7 @@ fn run_worker(command_rx: Receiver<IoPacket>) {
 
     let (submitter, mut submit_queue, mut complete_queue) = ring.split();
     let mut stats = Stats::new();
+    let mut retries = VecDeque::<IoPacket>::new();
 
     loop {
         stats.log();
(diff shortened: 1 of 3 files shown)


gabriele-0201:
* **#298** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/298?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

nomt/src/io/linux.rs
```diff
@@ -88,17 +90,24 @@ fn run_worker(command_rx: Receiver<IoCommand>, handle_tx: Vec<Sender<CompleteIo>
...
-                    Err(std::io::Error::from_raw_os_error(completion_event.result()))
-                } else {
-                    Ok(())
+                let result = match command.kind.get_result(completion_event.result()) {
+                    IoKindResult::Ok => Ok(()),
+                    IoKindResult::Err => {
+                        Err(std::io::Error::from_raw_os_error(completion_event.result()))
+                    }
+                    IoKindResult::Retry => {
+                        retries.push_back(command);
```
deuszx, line 105: Is it true that a write can be pushed back and the end of the queue for a retry? Wouldn't you possibly change the order of writes, which might lead to the last write being an incorrect one?

gabriele-0201: The io module does not guarantee the order in which commands are performed, this is an inherent property of io_uring.  It is up to the user to send commands, wait for completion, and properly interleave them with fsyncs

Nomt's workflow is defined within sessions, which involve reads, writes, and then fsyncs. (this is a generalization, there are multiple inner modules that adhere to a similar workflow)

deuszx: Thanks, I didn't realise there's no guaranteed order. Resolving.

rphmeier:
### Merge activity

* **Aug 22, 12:43 PM EDT**: @rphmeier started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/298).

PR #308, opened 2024-08-22
io: separate fsync from io_uring

gabriele-0201:
From https://kernel.dk/io_uring.pdf, in the section 'POLLED IO,' we can read: "Only op-codes that make sense for polled completion may be used on an io_uring instance registered with IORING_SETUP_IOPOLL" and fsync is not a valid opcode due to it being a blocking syscall.

This solves the problem of performing each fsync with the standard syscall without passing through io_uring.

This PR also fixes how the meta file was opened (without `O_DIRECT`) and corrects the usage of the metric PageCacheMisses.

### DIFF
--- a/nomt/src/io/mod.rs
+++ b/nomt/src/io/mod.rs
@@ -45,7 +45,6 @@ pub enum IoKind {
     Read(RawFd, u64, Box<Page>),
     Write(RawFd, u64, Box<Page>),
     WriteRaw(RawFd, u64, *const u8, usize),
-    Fsync(RawFd),
 }
 
 pub enum IoKindResult {
@@ -59,20 +58,11 @@ impl IoKind {
         match self {
             IoKind::Read(_, _, buf) | IoKind::Write(_, _, buf) => buf,
             IoKind::WriteRaw(_, _, _, _) => panic!("attempted to extract buf from write_raw"),
-            IoKind::Fsync(_) => panic!("attempted to extract buf from fsync"),
         }
     }
 
     pub fn get_result(&self, res: i32) -> IoKindResult {
         match self {
-            // fsync returns either -1 on failure or 0 on success
-            IoKind::Fsync(_) if res == 0 => IoKindResult::Ok,
-            // pread and pwrite return the number of bytes read or written
-            IoKind::Read(_, _, _) | IoKind::Write(_, _, _) | IoKind::Write(_, _, _)
-                if res == PAGE_SIZE as i32 =>
-            {
-                IoKindResult::Ok
-            }
             // pread returns 0 if the file has been read till the end of file
             //
             // This could be a failure because the end of the file could be smaller than PAGE_SIZE.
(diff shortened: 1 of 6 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/308?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#308** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/308?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#298** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/298?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

nomt/src/store/writeout.rs
```diff
@@ -404,15 +384,16 @@ impl MetaSwap {
...
-                return false;
             }
         }
 
         if self.should_fsync {
-            if let Err(_) = io.try_send_meta(IoKind::Fsync(self.meta_fd)) {
-                return false;
+            unsafe {
+                let f = File::from_raw_fd(self.meta_fd);
+                f.sync_all().unwrap();
```
rphmeier, line 393: it's cleaner to just call libc::fsync here

nomt/src/io/mod.rs
```diff
@@ -59,27 +58,20 @@ impl IoKind {
...
             // pread returns 0 if the file has been read till the end of file
             //
             // This could be a failure because the end of the file could be smaller than PAGE_SIZE.
             // However, as each read operation follows a write operation,
             // there should be no unexpected end-of-file that is not aligned with PAGE_SIZE
             // when all previous writes have succeeded.
             IoKind::Read(_, _, _) if res == 0 => IoKindResult::Ok,
+            // pread and pwrite return the number of bytes read or written
+            _ if res == PAGE_SIZE as i32 => IoKindResult::Ok,
             _ if res == -1 => IoKindResult::Err,
```
rphmeier, line 75: Another thing for a follow-up:

> res holds the result of the request. Think of it like the return
value from a system call. For a normal read/write operation, this will be like the return value from read(2) or
write(2). For a successful operation, it will contain the number of bytes transferred. If a failure occurred, it will
contain the negative error value. For example, if an I/O error occurred, res will contain -EIO.

Checking for `res == -1` doesn't seem right (as you've seen in debugging this issue). Instead, we need to handle `res < 0` and turn `-res` into a std::io::Error.

rphmeier (approved):

nomt/src/io/mod.rs
gabriele-0201: The conversion is already being done in the `linux.rs`. This method uses syscall results to make it shareable with `mod.rs`

nomt/src/store/writeout.rs
gabriele-0201: Fair! thanks! I will update it right now!

rphmeier (approved):

rphmeier:
### Merge activity

* **Aug 22, 12:43 PM EDT**: @rphmeier started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/308).

PR #323, opened 2024-08-24
bitbox sync sync

pepyakin:


pepyakin:
* **#323** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/323?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

nomt/src/bitbox/wal/mod.rs
```diff
@@ -1,243 +1,132 @@
...
-        // file pool. FALLOC_FL_COLLAPSE_RANGE only works on Linux.
-        #[cfg(target_os = "linux")]
-        unsafe {
-            let res = libc::fallocate(
-                self.wal_file.as_raw_fd(),
-                libc::FALLOC_FL_COLLAPSE_RANGE,
+impl Mmap {
+    fn new(sz: usize) -> Self {
+        let ptr = unsafe {
+            libc::mmap(
```
rphmeier, line 12: this seems correct as a way to allocate the memory needed for the WAL, but I wonder if mmap is overkill here

nomt/src/bitbox/wal/mod.rs
```diff
@@ -1,243 +1,132 @@
...
-
-// Open a WAL file and append encoded batch of writes to the end,
-// and prune previous ones if the current succeeded.
-//
-// It does not perform any check on the state of the WAL file
-pub struct WalWriter {
-    wal_file: File,
+struct Mmap {
+    ptr: *mut u8,
+    sz: usize,
```
rphmeier, line 6: nit: `size` is clearer than `sz`

nomt/src/store/writeout.rs
```diff
@@ -212,11 +307,83 @@ impl IoDmux {
...
+    fn send_writes(&mut self, io: &mut IoDmux) {
+        // send all writes
+        while self.wal_blob_pns.len() > 0 {
+            // UNWRAP: we know the list is not empty.
+            let pn = self.wal_blob_pns.pop().unwrap();
+
+            unsafe {
+                let ptr = self.wal_blob.0.add(pn as usize * PAGE_SIZE);
+                let len = PAGE_SIZE;
+                if let Err(_) = io.try_send_wal(IoKind::WriteRaw(self.wal_fd, pn as u64, ptr, len))
```
rphmeier, line 359: I'm not sure this is OK, because there are no guarantees that io_uring executes requests in order.
There are two things I think need to be true for this to be correct:
  1. That `pwrite` doesn't care about the actual offset when an fd is opened with  `O_APPEND` and always appends the page.
  2. That the WAL pages are ordering independent.
  
  
Quoting from https://linux.die.net/man/2/pwrite

> POSIX requires that opening a file with the O_APPEND flag should have no affect on the location at which pwrite() writes data. However, on Linux, if a file is opened with O_APPEND, pwrite() appends data to the end of the file, regardless of the value of offset. 

It seems that (1) is possibly guaranteed on linux but that this behavior can't be relied on for other POSIXes.

I see the need for a hack and understand the issue this is solving, but it seems better to just avoid the `io` subsystem altogether and do the write in a background thread or something.

nomt/src/bitbox/wal/mod.rs
pepyakin: I guess as it stands right now — yes. This was written for the single blob case but now we still split the whole thing into pages in the writeout.

that being said, i still hope we come back to the single blob approach and changing this now would be extra effort.

nomt/src/store/writeout.rs
pepyakin: It's true that the pages will land in no particular order. However, given that they all land, the file will have the same contents regardless of the order they actually got written. The requirement for a successful sync is that the manifest gets updated and that only happens iff WAL is successfuly fsynced.

rphmeier: I want to be clear: that is only true if 

> 1. pwrite doesn't care about the actual offset when an fd is opened with O_APPEND and always appends the page

nomt/src/bitbox/wal/mod.rs
rphmeier: yes, that definitely makes sense. We should discuss how to return to a single blob approach. I don't view that as a blocker for this PR.

nomt/src/bitbox/mod.rs
```diff
@@ -158,85 +114,39 @@ impl DB {
...
-                        page_id: page_id.encode(),
-                        changed: get_changed(&page, &page_diff),
-                        page_diff: page_diff.get_raw(),
-                        bucket_index: bucket as u64,
-                    });
+                    let changed = get_changed(&page, &page_diff);
+                    wal_blob_builder.write_update(
+                        page_id.encode(),
+                        page_diff.get_raw(),
+                        changed,
```
rphmeier, line 121: noting an optimization possibility here (for a later PR)

the impact of memory allocation in `get_changed` is not insignificant. altering `wal_blob_builder.write_update` to accept an iterator as an argument would avoid that.

nomt/src/store/writeout.rs
pepyakin: Ah, my bad! I was sure I committed the changes that removed O_APPEND but apparently somehow missed those.

PR #337, opened 2024-08-26
opt: use aligned page in page cache and avoid some copies

rphmeier:
By switching page loads from `Vec<u8>` to aligned pages, I've eliminated all but one copy during page cache commit and hash table writes. This last copy can be elided by having writeout draw directly from the page cache, but that's a large piece of work that will require some discussion before diving into.

### DIFF
--- a/nomt/src/page_cache.rs
+++ b/nomt/src/page_cache.rs
@@ -1,5 +1,6 @@
 use crate::{
     bitbox::BucketIndex,
+    io::Page as AlignedPage,
     metrics::{Metric, Metrics},
     page_region::PageRegion,
     rw_pass_cell::{ReadPass, Region, RegionContains, RwPassCell, RwPassDomain, WritePass},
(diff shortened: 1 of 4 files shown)


rphmeier:
* **#337** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/337?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

nomt/src/page_cache.rs
```diff
@@ -1,5 +1,6 @@
 use crate::{
     bitbox::BucketIndex,
+    io::Page as AlignedPage,
```
pepyakin, line 3: Nit: not super excited about the name since that's a bit of tautology. If you think about it, pages are always aligned, that's kinda whole idea. 

To disambiguate I suggest just using `io::Page`: it's more concise, it's unambigious, it's a bit more greppable.

PR #342, opened 2024-08-26
opt: use Select to avoid busy-looping in warm up

rphmeier:
I had noticed that, due to the way `Seek` was written, warmup workers tended to use a full core.

This breaks up `Seeker::advance` into its constituent parts and has warm-up workers use a `Select` to avoid wasting CPU. I observed this using a lot less CPU now (20% in benchtop vs 100%)

Note that we do pay a cost here, though the effect wasn't measurably worse from what I saw. The cost is that sending warm-ups to the worker now has to wake up the thread, adding some overhead to reads and writes.

### DIFF
--- a/nomt/src/bitbox/mod.rs
+++ b/nomt/src/bitbox/mod.rs
@@ -292,6 +292,11 @@ impl PageLoader {
             Err(_) => anyhow::bail!("I/O pool hangup"),
         }
     }
+
+    /// Get the underlying I/O handle.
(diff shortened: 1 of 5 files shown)


rphmeier:
* **#342** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/342?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

PR #350, opened 2024-08-27
beatree separate bench

gabriele-0201:
~~I couldn't find a better way to expose the private functions that need to be benchmarked besides starting to use this feature flag, which requires launching benchmarks with the following command: `cargo bench --features benchmarks`. Another problem is that I had to define the functions as `pub`~~

~~Do you have any other suggestions for handling this?~~

Update: To expose the benchmarks, there are currently three parts: the benchmarks implemented for the private function inside the same file, then a module `benches` inside the `beatree` module, and finally the `/benches/beatree.rs`, which simply calls the benchmark exposed by `beatree::benches`. This erases almost completely the code pollution of conditional compilation but requires a lot of modules to be defined as `pub(crate)` and the `beatree` module as `pub mod`. This is because the benches can only access public elements, thus there must be a chain of public modules to reach the final private function

### DIFF
--- a/nomt/Cargo.toml
+++ b/nomt/Cargo.toml
@@ -28,6 +28,7 @@ im = "15.1.0"
 lru = "0.12.3"
 libc = "0.2.155"
 itertools = "0.13.0"
+criterion = { version = "0.3", optional = true }
 
 [target.'cfg(target_os="linux")'.dependencies]
 io-uring = "0.6.4"
@@ -39,6 +40,14 @@ loom = { version = "0.7", features = ["checkpoint"] }
 rand_pcg = "0.3.1"
 hex-literal = "0.4"
 tempfile = "3.8.1"
+criterion = "0.3"
 
 [lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(loom)'] }
+
+[[bench]]
+name = "beatree"
+harness = false
+
+[features]
+benchmarks = ["dep:criterion"]
--- a/nomt/benches/beatree.rs
+++ b/nomt/benches/beatree.rs
@@ -1,0 +1,7 @@
+#![cfg(feature = "benchmarks")]
+
+use criterion::{criterion_group, criterion_main};
+use nomt::beatree::benches::beatree_benchmark;
+
+criterion_group!(benches, beatree_benchmark);
+criterion_main!(benches);
--- a/nomt/src/beatree/ops/update/mod.rs
+++ b/nomt/src/beatree/ops/update/mod.rs
@@ -21,7 +21,7 @@ use branch::{BaseBranch, BranchUpdater, DigestResult as BranchDigestResult};
 use leaf::{BaseLeaf, DigestResult as LeafDigestResult, LeafUpdater};
 
 mod branch;
-mod leaf;
+pub(crate) mod leaf;
 
 // All nodes less than this body size will be merged with a neighboring node.
 const BRANCH_MERGE_THRESHOLD: usize = BRANCH_NODE_BODY_SIZE / 2;
--- a/nomt/src/lib.rs
+++ b/nomt/src/lib.rs
@@ -26,7 +26,13 @@ pub use nomt_core::proof;
 pub use nomt_core::trie::{KeyPath, LeafData, Node};
 pub use options::Options;
 
+// beatree module needs to be exposed to be benchmarked
+#[cfg(feature = "benchmarks")]
+#[allow(missing_docs)]
+pub mod beatree;
+#[cfg(not(feature = "benchmarks"))]
 mod beatree;
+
 mod bitbox;
 mod commit;
 mod metrics;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -76,6 +76,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
 
 [[package]]
+name = "atty"
+version = "0.2.14"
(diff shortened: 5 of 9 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/350?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#351** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/351?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#350** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/350?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#349** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/349?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

nomt/src/beatree/ops/update/mod.rs
```diff
@@ -19,6 +19,9 @@ use super::BranchId;
 use branch::{BaseBranch, BranchUpdater, DigestResult as BranchDigestResult};
 use leaf::{BaseLeaf, DigestResult as LeafDigestResult, LeafUpdater};
 
+#[cfg(feature = "benchmarks")]
+pub use leaf::separate;
```
rphmeier, line 23: maybe find another way to do this. exporting this all from the global namespace is a pretty big code smell.

nomt/benches/beatree.rs
```diff
@@ -1,7 +1,41 @@
...
+                b.iter(|| separate(&input.0, &input.1));
+            },
+        );
+    }
+
+    group.finish();
+}
+
+pub fn beatreee_benchmark(c: &mut Criterion) {
+    separate_benchmark(c);
```
rphmeier, line 38: It would be better to have multiple benchmarks, each for their own number of shared _bits_

nomt/benches/beatree.rs
```diff
@@ -1,7 +1,41 @@
...
+    b[shared_bytes] = 1;
+
+    (a, b)
+}
+
+fn separate_benchmark(c: &mut Criterion) {
+    let mut group = c.benchmark_group("separate");
+
+    for shared_bytes in 0..32 {
+        let input = get_keys(shared_bytes);
```
rphmeier, line 24: better not to include generating keys in a benchmark, especially a microbenchmark. generate the data up-front, then benchmark it.

nomt/benches/beatree.rs
gabriele-0201: It is already set up like this: having a benchmarking group and using `bench_with_input` results in multiple benchmarks. Currently, there are shared bytes ranging from 0 to 32 (I will reduce the range to what you described, as they otherwise take too long)

or do you mean something different?

an example of `cargo bench --features benchmarks` is:

```shell
separate/0              time:   [85.400 ns 85.514 ns 85.683 ns]
                        change: [+42.966% +44.113% +45.865%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 24 outliers among 100 measurements (24.00%)
  8 (8.00%) low mild
  5 (5.00%) high mild
  11 (11.00%) high severe
separate/1              time:   [127.01 ns 127.16 ns 127.42 ns]
                        change: [-0.0945% -0.0070% +0.1108%] (p = 0.91 > 0.05)
                        No change in performance detected.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) high mild
  8 (8.00%) high severe
separate/2              time:   [167.66 ns 168.08 ns 168.73 ns]
                        change: [-5.0945% -4.2770% -3.5234%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe
```

nomt/src/beatree/ops/update/mod.rs
gabriele-0201: It doesn't seem to exist a perfect solution. The main problem is that there are private functions inside private modules that need to be benchmarked by something outside of the crate. The only drastic solution I see is to start using `#[cfg(bench)]`, which is still a nightly feature.

I hope that what is proposed in the new update of the PR is ok for you!

nomt/Cargo.toml
```diff
@@ -28,6 +28,7 @@ im = "15.1.0"
 lru = "0.12.3"
 libc = "0.2.155"
 itertools = "0.13.0"
+criterion = {version = "0.3", optional = true }
```
pepyakin, line 31:
```suggestion
criterion = { version = "0.3", optional = true }
```

nomt/src/lib.rs
```diff
@@ -26,7 +26,7 @@ pub use nomt_core::proof;
 pub use nomt_core::trie::{KeyPath, LeafData, Node};
 pub use options::Options;
 
-mod beatree;
+pub mod beatree;
```
pepyakin, line 29: Ideally, we don't expose this to users. If there is absolutely no way, a comment should explain why there is this internal part sticking out.

gabriele-0201: Yes, I will add a comment for sure. Also, thinking about exposing this module to the user is not great at all. So, I will find a way to put at least one conditional compilation here to expose it only during benchmarks

rphmeier (approved):
Yep this looks better

rphmeier:
### Merge activity

* **Aug 29, 5:53 AM EDT**: @rphmeier merged this pull request with [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/350).

PR #355, opened 2024-08-27
opt: avoid hashmap in page walker page diffs

rphmeier:
This doesn't currently show up in profiles in overall performance, since so much time is wasted in the seeker, but anecdotally this caused PageWalker::advance_and_replace to now 21% of its time hashing, instead of the previous 7.5%.

### DIFF
--- a/nomt/src/page_walker.rs
+++ b/nomt/src/page_walker.rs
@@ -49,7 +49,6 @@ use nomt_core::{
     trie::{self, KeyPath, Node, NodeHasher, NodeHasherExt, NodeKind, ValueHash},
     trie_pos::TriePosition,
 };
-use std::collections::HashMap;
 
 use crate::{
     page_cache::{Page, PageCache, PageDiff, ShardIndex},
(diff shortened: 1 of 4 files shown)


rphmeier:
* **#355** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/355?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

deuszx:
Question for the sake of learning - how do you decide to switch from HashMap to Vec if you're saying that it doesn't show up in benchmarks. Intuition?

nomt/src/page_walker.rs
```diff
@@ -69,12 +68,18 @@ pub enum Output {
...
     ///
     /// This is always the output when a parent page is supplied to the walker.
-    ChildPageRoots(Vec<(TriePosition, Node)>, HashMap<PageId, PageDiff>),
+    ChildPageRoots(Vec<(TriePosition, Node)>, Vec<(PageId, PageDiff)>),
+}
+
+struct StackItem {
+    page_id: PageId,
+    page: Page,
+    diff: PageDiff,
```
pepyakin, line 82: Nit: the diff was added but it's not immediatelly obvious why. While it's easy to see why we have the two former items, it's not so easy with `diff`.

rphmeier:
> Question for the sake of learning - how do you decide to switch from HashMap to Vec if you're saying that it doesn't show up in benchmarks. Intuition?

What I mean in particular is that it shows up in the part of the profile relevant to this code, so I can see the 'micro' effect. But it doesn't improve the overall runtime of the process.

nomt/src/page_walker.rs
rphmeier: IMO it's somewhat clear from the context of the page walker.

We walk the tree of pages from left to right, which means that our stack at any point is a kind of cross-section of the page tree. While pages are in that stack, they change. When they leave the stack, they won't be edited any more and we can save their diff.

(note: there are some edge cases where we pop a page off the stack and then right afterwards we either edit it or push it back for a single operation, and that is because of `write_leaf_children`, but in general this principle holds)

PR #369, opened 2024-08-29
feat: closed form for leaf overflow total_needed_pages

gabriele-0201:
The function could be made smaller if the second condition (`needed_pages_raw_value <= MAX_OVERFLOW_CELL_NODE_POINTERS + available_page_numbers`) is removed and the latter formula is performed using f64. I didn't like casting everything to f64, so I found a way to make it work with integers


rphmeier (approved):
Yep, this makes sense!

rphmeier:
> I didn't like casting everything to f64, so I found a way to make it work with integers

Yes, FP should be avoided in precision code

PR #374, opened 2024-08-30
fix: simplify writeout to block on send / recv

rphmeier:
The code here had some issues with busy-looping on `try_send` and `try_recv`.

This actually caused livelocks, since we were calling `try_recv_*` when we failed to submit a write. The sequence of events would be this:
  1. There would be no writes for e.g. the LN pending in the system but the I/O queue would be full.
  2. We'd fail to submit the next LN write.
  3. We'd then busy-loop on the next LN completion, which would never come because of (1).
   
It doesn't seem necessary to do this as a state machine per se. I might imagine that the inboxes could get pretty full, but we don't actually use the results and could easily just track the number of received items.

### DIFF
--- a/nomt/src/store/writeout.rs
+++ b/nomt/src/store/writeout.rs
@@ -4,8 +4,6 @@ use std::{
     os::fd::{FromRawFd, IntoRawFd, RawFd},
 };
 
-use crossbeam_channel::TrySendError;
-
 use crate::io::{CompleteIo, IoCommand, IoHandle, IoKind, Page, PAGE_SIZE};
 
 use super::{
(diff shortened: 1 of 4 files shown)


rphmeier:
* **#374** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/374?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

nomt/src/store/writeout.rs
```diff
@@ -52,11 +50,7 @@ pub fn run(
                 ht_remaining: ht.len(),
                 ht,
             },
-            meta_swap: Some(MetaSwap {
-                meta_fd,
-                new_meta: Some(new_meta),
-                should_fsync: false,
-            }),
+            meta_swap: Some(MetaSwap { meta_fd, new_meta }),
```
gabriele-0201, line 53: Option on `meta_swap` seems to be no longer needed

gabriele-0201 (approved):

rphmeier:
### Merge activity

* **Sep 2, 11:01 AM EDT**: @rphmeier started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/374).

PR #383, opened 2024-09-04
feat: new branch node encoding

gabriele-0201:
stop using a common prefix_len and instead use a vector of cells
to store different prefix_len values, one for each separator

### DIFF
--- a/nomt/src/beatree/branch/node.rs
+++ b/nomt/src/beatree/branch/node.rs
@@ -7,17 +7,17 @@ use crate::beatree::Key;
 // Here is the layout of a branch node:
 //
 // ```rust,ignore
-// bbn_pn: u32            // the page number this is stored under,
-//                        // if this is a BBN, 0 otherwise.
+// bbn_pn: u32          // the page number this is stored under,
+//                      // if this is a BBN, 0 otherwise.
 //
-// n: u16                 // item count
-// prefix_len: u16         // bits
-// separator_len: u16      // bits
+// n: u16               // item count
+// prefix_len: u16      // bits
+// cells: u16[n]        // bit offsets of the end of separators within the separators bitvec
 //
-// # Then the varbits follow. To avoid two padding bytes, the varbits are stored in a single bitvec.
+// # To avoid two padding bytes, prefix and separators are stored in a single bitvec.
 //
 // prefix: bitvec[prefix_len]
-// separators: bitvec[n * separator_len]
+// separators: bitvec
 //
 // # Node pointers follow. The list is aligned to the end of the node, with the last item in the
 // # list occupying the last 4 bytes of the node.
@@ -25,7 +25,7 @@ use crate::beatree::Key;
 // node_pointers: LNPN or BNID[n]
 // ```
 
-const BRANCH_NODE_HEADER_SIZE: usize = 4 + 2 + 2 + 2;
+const BRANCH_NODE_HEADER_SIZE: usize = 4 + 2 + 2;
 pub const BRANCH_NODE_BODY_SIZE: usize = BRANCH_NODE_SIZE - BRANCH_NODE_HEADER_SIZE;
 
 /// A branch node, regardless of its level.
@@ -77,30 +77,37 @@ impl BranchNode {
         slice[6..8].copy_from_slice(&len.to_le_bytes());
     }
 
-    pub fn separator_len(&self) -> u16 {
-        self.view().separator_len()
-    }
-
-    pub fn set_separator_len(&mut self, len: u16) {
-        let slice = self.as_mut_slice();
-        slice[8..10].copy_from_slice(&len.to_le_bytes());
-    }
-
-    fn varbits_mut(&mut self) -> &mut BitSlice<u8, Msb0> {
-        let body_end = body_size(
-            self.prefix_len() as _,
-            self.separator_len() as _,
-            self.n() as _,
-        ) + BRANCH_NODE_HEADER_SIZE;
-        self.as_mut_slice()[BRANCH_NODE_HEADER_SIZE..body_end].view_bits_mut()
-    }
-
     pub fn prefix(&self) -> &BitSlice<u8, Msb0> {
         self.view().prefix()
     }
 
+    fn set_prefix(&mut self, prefix: &BitSlice<u8, Msb0>) {
+        let start = BRANCH_NODE_HEADER_SIZE + self.n() as usize * 2;
+        let prefix_len = self.prefix_len() as usize;
+        self.as_mut_slice()[start..].view_bits_mut()[..prefix_len].copy_from_bitslice(prefix);
+    }
+
     pub fn separator(&self, i: usize) -> &BitSlice<u8, Msb0> {
         self.view().separator(i)
+    }
+
+    fn set_separator(
+        &mut self,
+        i: usize,
+        separator: &BitSlice<u8, Msb0>,
+        bit_offset_start: usize,
+        bit_offset_end: usize,
+    ) {
+        let n = self.n() as usize;
+        let prefix_len = self.prefix_len() as usize;
+        let slice = self.as_mut_slice();
+
+        let cells_start = BRANCH_NODE_HEADER_SIZE + (i * 2);
+        slice[cells_start..][..2].copy_from_slice(&(bit_offset_end as u16).to_le_bytes());
+
+        let separators_start = BRANCH_NODE_HEADER_SIZE + (n * 2);
+        slice[separators_start..].view_bits_mut()[prefix_len..][bit_offset_start..bit_offset_end]
+            .copy_from_bitslice(&separator);
     }
 
     pub fn node_pointer(&self, i: usize) -> u32 {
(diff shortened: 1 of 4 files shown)


gabriele-0201:
* **#383** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/383?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

nomt/src/beatree/branch/node.rs
```diff
@@ -173,61 +169,63 @@ impl<'a> BranchNodeView<'a> {
...
+    // up to the next byte boundary. They are preceded by cells and followed by node pointers
+    (n + 1) * 2 + (prefix_len + tot_separators_len + 7) / 8 + (n * 4)
 }
 
 pub struct BranchNodeBuilder {
     branch: BranchNode,
     index: usize,
     prefix_len: usize,
-    separator_len: usize,
+    cells: Vec<u8>,
```
rphmeier, line 182: I don't think we should be allocating memory inside of the branch node builder, either here or with `separators`. 

You might want to look at the leaf builder / leaf gauge to see how the API might look for this.
With `BranchNodeBuilder::new(branch, n, prefix_len, sum_of_separator_lens)` you should be able to encode without doing any intermediate allocations.

nomt/src/beatree/ops/update/branch.rs
```diff
@@ -406,8 +406,7 @@ impl BranchUpdater {
 struct BranchGauge {
     first_separator: Option<Key>,
     prefix_len: usize,
-    // total separator len, not counting prefix.
-    separator_len: usize,
+    separators_len: Vec<usize>,
```
rphmeier, line 409: related: https://github.com/thrumdev/nomt/pull/383/files#r1743939784

you don't need a `Vec` here.

also, small grammar nit: it's more accurate to name this field `separator_lengths` (though I expect this field will go away)

nomt/src/beatree/ops/update/branch.rs
```diff
@@ -416,46 +415,56 @@ impl BranchGauge {
...
+            self.separators_len.push(self.prefix_len);
+        }
+
+        self.separators_len.push(len);
+
+        self.prefix_len = prefix_len(first, &key);
         self.n += 1;
     }
 
+    fn tot_separators_len(&self) -> usize {
```
rphmeier, line 445: style: don't abbreviate

nomt/src/beatree/branch/node.rs
```diff
@@ -173,61 +169,63 @@ impl<'a> BranchNodeView<'a> {
 
 unsafe impl Send for BranchNode {}
 
-pub fn body_size(prefix_len: usize, separator_len: usize, n: usize) -> usize {
+pub fn body_size(prefix_len: usize, tot_separators_len: usize, n: usize) -> usize {
```
rphmeier, line 172: style: don't abbreviate `total` to `tot`.

nomt/src/beatree/branch/node.rs
```diff
@@ -77,30 +80,31 @@ impl BranchNode {
...
-        ) + BRANCH_NODE_HEADER_SIZE;
-        self.as_mut_slice()[BRANCH_NODE_HEADER_SIZE..body_end].view_bits_mut()
+    pub fn separator(&self, i: usize) -> &BitSlice<u8, Msb0> {
+        self.view().separator(i)
     }
 
-    pub fn prefix(&self) -> &BitSlice<u8, Msb0> {
-        self.view().prefix()
-    }
+    fn set_separators(&mut self, cells: Vec<u8>, separators: BitVec<u8, Msb0>) {
```
rphmeier, line 97: this also seems unnecessary if you can get rid of the allocations in `BranchBuilder`.

nomt/src/beatree/branch/node.rs
gabriele-0201: > Altering the gauge to produce sum_of_separator_lens_not_counting_prefix should be pretty straightforward. First you sum sum_of_separator_lens (not applying prefix compression). Then you subtract prefix_len * n from that value when calculating the body size.

I don't think this could work (given `sum_of_separator_lens` being gathered during ingestion in each gauge.ingest step)

The prefix_len could be bigger than some separator_len specified in `branch_gauge.ingest`, thus resulting in `sum_of_separator_lens - prefix_len * n` yielding fewer bytes than actually occupied.

I don't see how it could work without using something to store the separator lengths.
We could overestimate the total occupied bytes by simply summing `max(prefix_len, separator_len)` at each ingestion step.
However, the `prefix_len` may decrease over time, resulting in a size larger than the actual space that will be occupied in the branch node.

gabriele-0201: > Altering the gauge to produce sum_of_separator_lens_not_counting_prefix should be pretty straightforward. First you sum sum_of_separator_lens (not applying prefix compression). Then you subtract prefix_len * n from that value when calculating the body size.

I don't think this could work (given `sum_of_separator_lens` being gathered during ingestion in each gauge.ingest step)

The prefix_len could be bigger than some separator_len specified in `branch_gauge.ingest`, thus resulting in `sum_of_separator_lens - prefix_len * n` yielding fewer bytes than actually occupied.

I don't see how it could work without using something to store the separator lengths.
We could overestimate the total occupied bytes by simply summing `max(prefix_len, separator_len)` at each ingestion step.
However, the `prefix_len` may decrease over time, resulting in a size larger than the actual space that will be occupied in the branch node.

I tested this by making a small change to `tot_separators_len`

```rust
    fn tot_separators_len(&self) -> usize {
        let r1 = self
            .separators_len
            .iter()
            .map(|s| s.saturating_sub(self.prefix_len))
            .sum();
        let sum_of_separator_lens_not_counting_prefix = self.separators_len.iter().sum::<usize>();
        let r2 =  sum_of_separator_lens_not_counting_prefix - (self.prefix_len * self.n);
        if self.n != 1 {
            assert_eq!(r1, r2); // this fails immediately 
        }
        r1
    }
```

In the first iteration of `cargo run --release -- run -w transfer -c 20 -b nomt --time-limit 60s -s 10000 -r --buckets 500000` with `n=2` we have `&self.separators_len = [1, 17,]` and `self.prefix_len = 13`, which results in a negative outcome in this case

nomt/src/beatree/branch/node.rs
```diff
@@ -7,25 +7,28 @@ use crate::beatree::Key;
...
 //
 // # Node pointers follow. The list is aligned to the end of the node, with the last item in the
 // # list occupying the last 4 bytes of the node.
 //
 // node_pointers: LNPN or BNID[n]
 // ```
 
-const BRANCH_NODE_HEADER_SIZE: usize = 4 + 2 + 2 + 2;
+const BRANCH_NODE_HEADER_SIZE: usize = 4 + 2 + 2;
+pub const BRANCH_NODE_EMPTY_BODY: usize = 2;
```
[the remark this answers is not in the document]
gabriele-0201: I folded #384 into this PR!

nomt/src/beatree/branch/node.rs
gabriele-0201: > Altering the gauge to produce sum_of_separator_lens_not_counting_prefix should be pretty straightforward. First you sum sum_of_separator_lens (not applying prefix compression). Then you subtract prefix_len * n from that value when calculating the body size.

I don't think this could work (given `sum_of_separator_lens` being gathered during ingestion in each gauge.ingest step)

The prefix_len could be bigger than some separator_len specified in `branch_gauge.ingest`, thus resulting in `sum_of_separator_lens - prefix_len * n` yielding fewer bytes than actually occupied.

I don't see how it could work without using something to store the separator lengths.

I tested this by making a small change to `tot_separators_len`

```rust
    fn tot_separators_len(&self) -> usize {
        let r1 = self
            .separators_len
            .iter()
            .map(|s| s.saturating_sub(self.prefix_len))
            .sum();
        let sum_of_separator_lens_not_counting_prefix = self.separators_len.iter().sum::<usize>();
        let r2 =  sum_of_separator_lens_not_counting_prefix - (self.prefix_len * self.n);
        if self.n != 1 {
            assert_eq!(r1, r2); // this fails immediately 
        }
        r1
    }
```

In the first iteration of `cargo run --release -- run -w transfer -c 20 -b nomt --time-limit 60s -s 10000 -r --buckets 500000` with `n=2` we have `&self.separators_len = [1, 17,]` and `self.prefix_len = 13`, which results in a negative outcome in this case

rphmeier: I see. I hadn't considered the case where prefix_len was greater than the separator len. I suppose that if we have two keys:

`10000...` (total separator length 1)
`100001...` (total separator length 6)

the prefix length will be 5.

> I don't see how it could work without using something to store the separator lengths.

I can understand how you arrived at this conclusion based off of this edge case. However, I think we should be able to overcome this behavior by setting `separator_len` to `max(prefix_len, separator_len)`. It is always safe to logically zero-extend separators. The encoding of the branch with the example above would be

`[prefix: 10000][separator 1: len 0|][separator 2: len 1|1]` 

This will still require something to store separator lengths, but hopefully we can just reuse the `ops` Vec in the branch updater for that purpose.

I don't view that as a blocker for this PR and I am OK with merging it as-is and addressing later.

rphmeier: The other option is to find a way to reuse the allocation. I don't think that copying is avoidable here. We just shouldn't be allocating memory on every branch we build!

rphmeier (approved):

nomt/src/beatree/branch/node.rs
gabriele-0201: >It's worth noting that only the very first key can have separator_len <= prefix_len.

Yes, that's absolutely true. It is not only worth noting, but it is the key point that enables us to address the problem effectively and cleanly without any allocation

> When you push a new item, the prefix length must either reduce or stay the same

The second push could cause the prefix_len to increase, from the third push onwards, it can only reduce or remain the same.
However, this is not a problem because we simply store the length of the first separator.
In fact, in my previous code, the issue was solely with the *first* separator length.

The following code works perfectly fine:

```rust
    fn total_separators_len(&self) -> usize {
        let r1 = self
            .separator_lens
            .iter()
            .map(|s| s.saturating_sub(self.prefix_len))
            .sum();
        let r2 = self.first_separator.unwrap().1 // first separator_len
            .saturating_sub(self.prefix_len)
            + self.sum_separator_lens - (self.n - 1) * self.prefix_len;
        if r1 != r2 {
            panic!();
        }
        r2
    }
```

where `sum_separator_lens` is increased each ingest step (beside the first one) in the following manner
`self.sum_separator_lens += separator_len;`

>When you push a new item, the prefix length must either reduce or stay the same. You can increase prefix_compressed_separator_lengths by prefix_bits_lost * prefix_compressed_items.

I don't understand what you mean by 'prefix_bits_lost' here.

Anyway, I just pushed updated code without allocations in the gauge :fire:  (the new code also contains a correction of the method body_size_after)

rphmeier: > I don't understand what you mean by 'prefix_bits_lost' here.

The change in the prefix len from pushing one key to the next

nomt/src/beatree/ops/update/branch.rs
```diff
@@ -416,44 +416,52 @@ impl BranchGauge {
...
         };
 
-        self.separator_len = std::cmp::max(len, self.separator_len);
-        self.prefix_len = std::cmp::min(self.separator_len, prefix_len(first, &key));
+        self.prefix_len = prefix_len(first, &key);
+        self.sum_separator_lens += len;
         self.n += 1;
     }
 
+    fn total_separators_len(&self, prefix_len: usize) -> usize {
```
rphmeier, line 438: nit: should be separator_lengths

nomt/src/beatree/ops/update/branch.rs
```diff
@@ -416,44 +416,52 @@ impl BranchGauge {
...
+        match self.first_separator {
+            Some((_, len)) => {
+                len.saturating_sub(prefix_len) + self.sum_separator_lens - (self.n - 1) * prefix_len
```
rphmeier, line 441: yep

rphmeier (approved):
Looks great now.

rphmeier:
### Merge activity

* **Sep 6, 6:43 PM EDT**: @rphmeier started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/383).

PR #396, opened 2024-09-05
refactor: page pool

pepyakin:
This changeset introduces a notion of page pool which changes the
approach we use to allocate the pages for IO. Instead of allocating
pages from the global allocator one-by-one, the page pool allocates a
big region of virtual memory from the system and slices it on pages
maintaining a freelist for efficient allocation.

This changeset is only a stepping stone and there is a bit of work to
do. Specifically, it introduces a concept of [`FatPage`] which is
relatively convenient to work with but at the cost of leaving some
performance on the table. Also, page pool's implementation is not the
most efficient out there. This is to be solved in future commits.

Suggestion for the review:

Most of the changes is plumbing: weaving the page pool into all the places page allocation happens.

### DIFF
--- a/nomt/src/io/linux.rs
+++ b/nomt/src/io/linux.rs
@@ -1,4 +1,4 @@
-use super::{CompleteIo, IoCommand, IoKind, IoKindResult, IoPacket, PAGE_SIZE};
+use super::{CompleteIo, IoCommand, IoKind, IoKindResult, IoPacket, PagePool, PAGE_SIZE};
 use crossbeam_channel::{Receiver, Sender, TryRecvError};
 use io_uring::{cqueue, opcode, squeue, types, IoUring};
 use slab::Slab;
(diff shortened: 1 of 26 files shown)


pepyakin:
* **#396** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/396?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

nomt/src/io/page_pool.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    /// Allocates a new [`Page`] and fills it with zeroes.
+    pub fn alloc_zeroed(&self) -> Page {
+        let page = {
+            let mut freelist = self.inner.freelist.write();
+            if freelist.is_empty() {
+                self.grow(&mut *freelist)
+            } else {
+                freelist.pop().unwrap()
```
rphmeier, line 147: nit: UNWRAP comment would be appreciated even though obvious.

nomt/src/io/linux.rs
```diff
@@ -128,15 +132,15 @@ fn run_worker(command_rx: Receiver<IoPacket>) {
     }
 }
 
-fn submission_entry(command: &mut IoCommand) -> squeue::Entry {
+fn submission_entry(command: &mut IoCommand, page_pool: &PagePool) -> squeue::Entry {
     match command.kind {
-        IoKind::Read(fd, page_index, ref mut buf) => {
-            opcode::Read::new(types::Fd(fd), buf.as_mut_ptr(), PAGE_SIZE as u32)
+        IoKind::Read(fd, page_index, ref mut page) => {
+            opcode::Read::new(types::Fd(fd), page.as_mut_ptr(page_pool), PAGE_SIZE as u32)
```
rphmeier, line 138: just flagging: this takes a read lock under the hood.

nomt/src/io/page_pool.rs
```diff
@@ -0,0 +1,203 @@
...
+            }
+        };
+        unsafe {
```
rphmeier, line 150: comment on `UNSAFE` would be good too.

deuszx:
@pepyakin I know this is a long shot but maybe you'll find a moment to teach a fellow developer :)

I'm curious how you arrived at a refactor like that - it'd seem to me that efficient memory allocation is a job of kernel (or a different layer in the stack but Lower than in the app code itself). Is there any particular benchmarking/approach tool that you had used? An experience or a hunch maybe?

Thanks.

PR #399, opened 2024-09-06
refactor: free list page ref & mut

pepyakin:
Adds types that encapsulate manipulation of a free list page.

PR #400, opened 2024-09-06
opt: lockless page pool

pepyakin:
This optimization changes the data structure by leveraging the fact that the regions are immutable once they were created. This allows us to avoid expensive locking after each access. The title is a bit of click-bait since locking still happens for freelist updates.

### DIFF
--- a/nomt/src/io/page_pool.rs
+++ b/nomt/src/io/page_pool.rs
@@ -1,9 +1,11 @@
-
 use super::PAGE_SIZE;
-use parking_lot::RwLock;
+use parking_lot::{RwLock, RwLockWriteGuard};
 use std::{
     ops::{Deref, DerefMut},
-    sync::Arc,
+    sync::{
+        atomic::{AtomicPtr, AtomicU32, Ordering},
+        Arc,
+    },
 };
 
 // Region is 256 MiB. The choice is mostly arbitrary, but:
@@ -12,17 +14,34 @@ use std::{
 // 2. it's a multiple of 2 MiB which is the size of huge-page on x86-64 and aarch64.
 // 3. it fits 65k 4 KiB pages which requires 2 bytes to address making it nice round number.
 //
+// If we take 16 bits for the slot index, we are left with another 16 bit which we can use to address
+// 64K regions. That makes up a total of 16 TiB of memory, way beyond of the current hardware limits.
+// Because of that, we limit ourselves to 4096 regions which gives us only 1 TiB of memory.
 const REGION_SLOT_BITS: u32 = 16;
 const REGION_SLOT_MASK: u32 = (1 << REGION_SLOT_BITS) - 1;
-const REGION_SIZE: usize = (1 << REGION_SLOT_BITS) * PAGE_SIZE;
+const SLOTS_PER_REGION: usize = 1 << REGION_SLOT_BITS;
+const REGION_BYTE_SIZE: usize = SLOTS_PER_REGION * PAGE_SIZE;
+const REGION_COUNT: usize = 4096;
 
 #[derive(Clone, Copy)]
 struct PageIndex(u32);
 
 impl PageIndex {
+    /// Creates a new page index from a region and a slot.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the region index is less than `REGION_COUNT` and the slot index
+    /// is less than `REGION_SLOT_MASK`.
+    fn from_region_and_slot(region: u32, slot: u32) -> Self {
+        assert!(region < REGION_COUNT as u32);
+        assert!(slot < SLOTS_PER_REGION as u32);
+        Self((region << REGION_SLOT_BITS) | slot)
+    }
+
     /// Extracts the region index from the reference.
     fn region(&self) -> usize {
-        (self.0 >> 16) as usize
+        (self.0 >> REGION_SLOT_BITS) as usize
     }
 
     /// Extracts the slot index within the region.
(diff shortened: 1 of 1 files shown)


pepyakin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/400?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#400** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/400?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#399** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/399?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

nomt/src/io/page_pool.rs
```diff
@@ -113,17 +132,23 @@ pub struct PagePool {
 }
 
 struct Inner {
-    regions: RwLock<Vec<*mut libc::c_void>>,
+    //
```
rphmeier, line 135: comment?

pepyakin: oh, yeah. Will add.

nomt/src/io/page_pool.rs
```diff
@@ -158,44 +189,63 @@ impl PagePool {
...
                 std::ptr::null_mut(),
-                REGION_SIZE,
+                REGION_BYTE_SIZE,
                 libc::PROT_READ | libc::PROT_WRITE,
                 libc::MAP_PRIVATE | libc::MAP_ANONYMOUS,
                 /* fd */ -1,
                 /* offset */ 0,
             )
         };
+        assert!(!region_ptr.is_null());
```
rphmeier, line 206: can you elaborate a bit?

nomt/src/io/page_pool.rs
```diff
@@ -158,44 +189,63 @@ impl PagePool {
...
-        for slot in 0..=REGION_SLOT_MASK {
-            freelist.push(Page(PageIndex(slot | region)));
+        // Next, we need to store the region pointer in the regions array.
+        //
+        // We store the pointer in the regions array before incrementing n_regions. This is not
+        // strictly necessary, because the freelist is empty and no page can refer to the new region
+        // yet. Likewise, drop cannot happen during this operation. We still do it in this order
+        // to just err on the safe side and avoid any potential issues.
+        let region_ix = self.inner.n_regions.load(Ordering::Relaxed);
+        self.inner.regions[region_ix as usize].store(region_ptr as *mut u8, Ordering::Release);
```
rphmeier, line 218: `Ordering::Release` doesn't look correct here. Release indicates that we want all previous writes to propagate to the next `Acquirer`. I believe that the correct order would be to store this with `Relaxed` and then to do the fetch-add to n_regions with `Ordering::Release`.

nomt/src/io/page_pool.rs
```diff
@@ -158,44 +189,63 @@ impl PagePool {
...
-        freelist.pop().unwrap()
+        // UNWRAP: we know that the freelist is not empty, because we just filled it.
+        freelist_guard.pop().unwrap()
     }
 
     fn data_ptr(&self, page_index: PageIndex) -> *mut u8 {
-        let region_ptr = self.inner.regions.read()[page_index.region()];
-        unsafe { region_ptr.add(page_index.slot_index() * PAGE_SIZE) as *mut u8 }
+        let region = page_index.region();
+        assert!(region < self.inner.n_regions.load(Ordering::Relaxed) as usize);
```
rphmeier, line 234: Just like the other comment. What we want here is to load this with `Acquire` ordering and then load the region pointer with `Relaxed`.

nomt/src/io/page_pool.rs
```diff
@@ -158,44 +189,63 @@ impl PagePool {
...
-            freelist.push(Page(PageIndex(slot | region)));
+        // Next, we need to store the region pointer in the regions array.
+        //
+        // We store the pointer in the regions array before incrementing n_regions. This is not
+        // strictly necessary, because the freelist is empty and no page can refer to the new region
+        // yet. Likewise, drop cannot happen during this operation. We still do it in this order
+        // to just err on the safe side and avoid any potential issues.
+        let region_ix = self.inner.n_regions.load(Ordering::Relaxed);
+        self.inner.regions[region_ix as usize].store(region_ptr as *mut u8, Ordering::Release);
+        self.inner.n_regions.fetch_add(1, Ordering::Relaxed);
```
rphmeier, line 219: note: this is protected by the write guard on the free list but if it were lock-free we'd want to claim the region index with a CAS rather than read followed by fetch-add.

nomt/src/io/page_pool.rs
pepyakin: this is a bug actually, below this line we should check for the MAP_FAILED which is actually returned. This is more of a hint that whatever we store in the `regions` field is not null.

pepyakin: ah, and in turn, we want the region ptrs, the allocated ones anyway, be non-null.

Why? Because that compromises safety (in the sense of UB):

1. you don't want to construct a slice from a ptr which is null.
2. you don't want to compute offset from null.

Now, when I was looking for a proof, I wasn't able to find the exact doc line stating that you cannot create offsets against null. However, I just remember that from the experience. Here is the proof

https://godbolt.org/z/GnvsfbdPr

It might be possible that the compiler does it because it expects that each pointer points to a distinct "object" (for some definition of an object) and no other object could be reached through that pointer and null pointer is special, and in such a case where the compiler doesn't know that the ptr is statically always null then it won't pull out any UB.

But I think I spent way too much of my time even to write this message, so I would rather leave that check in place kek

nomt/src/io/page_pool.rs
pepyakin: yep, I don't think that matters much as we disccused and I also updated the comment to reflect that.

nomt/src/io/page_pool.rs
pepyakin: I suspect this is false. Orderings essentially specify how the memory outside of the atomic in question "synchronized", but the atomic itself is always "synchronized". That being said, I also feel that putting things in proper release-acquire brackets feels more right.

nomt/src/io/page_pool.rs
rphmeier: Thank you!

nomt/src/io/page_pool.rs
rphmeier: No, it's not false, when you account for compiler and CPU reordering. Proper acquire/release brackets are really important.

Atomicity guarantees operation ordering, nothing else. So in the first iteration that I commented on, it would be possible to read `n_regions=1` but `regions[1].load()` could be null. This could happen if the compiler or CPU reordered the stores, such that `n_regions` was incremented prior to storing the pointer in `regions[1]`, which was permitted by the `Release`.

Your mental model should be roughly that writes can move up across a release but not down below it. And that reads can move down across an acquire but not up above it.

PR #424, opened 2024-09-24
fix: prioritize completing single-page-request

gabriele-0201:
`take_completion` needs to prioritize single-page-requests and thus before taking completed requests it waits for the single-page-request before continuing with other requests

PR #422, opened 2024-09-23
refactor: leaf_stage range_extension handling

gabriele-0201:
the pr contains the following changes:
+ LeafChanges entries received from the right, if contains an inserted node, will always be used as new base
    Previously, if a new_entry had a bigger new_key, the base was searched through fresh leaves leading to a panic due to a double deletion of the leaf
+ `try_answer_left_neighbor` sends the right neighbor to the left one only if it has finished interacting with the right one
+ a small refactor inside `run_worker` on how the final calls to digest are handled 
+ a link to an issue related to the extension protocol

PR #421, opened 2024-09-20
fix: leaf_stage sort worker changes

gabriele-0201:
The branch_stage requires a sorted vector of changes. Thus, the leaf_stage needs to sort them when receiving the output of all workers

PR #420, opened 2024-09-20
fix: avoid removing cutoff from base leaf

gabriele-0201:
The cutoff needs to be removed only if all right neighbors have deleted all their leaves, and thus the current worker needs to create the new last leaf, which is done by removing the cutoff of the base leaf.

If the new_entry received from the request_range_extension is going to be the new base, it never requires removal of the cutoff. If the received entry is the last one, then new_entry.next_separator will already be None

PR #426, opened 2024-10-02
fix: extend range protocol supports ranges of untouched leaves

gabriele-0201:
extend range protocol now supports also workers sharing ranges of untouched leaves

The main changes are:

- `LeafChanges` becomes `LeavesTracker` because it tracks all the changes applied to leaves and which leaves are left untouched
- `ExtendRangeResponse` can now contain not only a new base to use but also information on whether the left worker needs to use a fresh leaf and how many of them need to be skipped

I'm not a huge fan of how I ended up handling `pending_base`. It feels misplaced to keep it inside `LeavesTracker`, but any other approach I thought would end up complicating the main `run_worker` function

Note: I am working on a possible optimization that uses a Vec instead of a BTreeMap inside the LeavesTracker

### DIFF
--- a/nomt/src/beatree/ops/update/leaf_stage.rs
+++ b/nomt/src/beatree/ops/update/leaf_stage.rs
@@ -22,7 +22,9 @@ use super::{
 };
 
 struct ExtendRangeResponse {
-    changed: Vec<(Key, ChangedLeafEntry)>,
+    deleted: Vec<(Key, ChangedLeafEntry)>,
+    new_base: Option<(Key, LeafNode, Option<Key>)>,
+    new_high_range: Option<Key>,
     new_right_neighbor: Option<Option<RightNeighbor>>,
 }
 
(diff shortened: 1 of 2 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/426?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#426** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/426?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#422** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/422?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#424](https://github.com/thrumdev/nomt/pull/424) <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/424?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#421** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/421?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#420** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/420?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#419** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/419?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

nomt/src/beatree/ops/update/leaf_stage.rs
```diff
@@ -22,7 +22,9 @@ use super::{
 };
 
 struct ExtendRangeResponse {
-    changed: Vec<(Key, ChangedLeafEntry)>,
+    deleted: Vec<(Key, ChangedLeafEntry)>,
+    new_base: Option<(Key, LeafNode, Option<Key>)>,
+    new_high_range: Option<Key>,
```
rphmeier, line 27: just having `changed` and `new_high` would have all the information necessary to proceed.

There are 3 possibilities:
  1. The next leaf is in `changed`.
  2. The next leaf is not in changed, but a higher leaf is in `changed`. therefore, the leaf hasn't changed
  3. The next leaf is not in changed, and there is no higher leaf in `changed`. without `new_high`, this previously led to an ambiguous situation that was the source of a bug.

It would be better to have this contain the minimal information possible.

nomt/src/beatree/ops/update/leaf_stage.rs
```diff
@@ -35,50 +37,95 @@ pub struct ChangedLeafEntry {
     pub inserted: Option<LeafNode>,
 
     // the separator of the next node.
-    pub next_separator: Option<Key>,
+    next_separator: Option<Key>,
+}
+
+pub enum LeavesTrackerEntry {
+    ChangedLeaf(ChangedLeafEntry),
+    UnchangedRange(SeparatorRange),
```
rphmeier, line 45: I'm ~ok with having this explicit, but IMO it is already implicit from the information in `changed`

nomt/src/beatree/ops/update/leaf_stage.rs
```diff
@@ -35,50 +37,95 @@ pub struct ChangedLeafEntry {
...
+            .entry(key)
+            .or_insert(LeavesTrackerEntry::ChangedLeaf(ChangedLeafEntry {
+                deleted: None,
+                inserted: None,
+                next_separator,
+            }));
+
+        let changed_entry = match entry {
+            LeavesTrackerEntry::ChangedLeaf(changed_entry) => changed_entry,
+            LeavesTrackerEntry::UnchangedRange(_) => panic!("TODO"),
```
rphmeier, line 69: & this is why I felt that the `UnchangedRange` type is kind of rough, and maybe better avoided. There are two issues with it:
  1. What key is appropriate for the `Map<Key, LeavesTrackerEntry>`? There is no perfect answer here
  2. Though it's impossible (bugs notwithstanding) this panic to occur, it does lead to slightly messy code, conceptually

nomt/src/beatree/ops/update/leaf_stage.rs
```diff
@@ -35,50 +37,95 @@ pub struct ChangedLeafEntry {
...
     }
 
-    pub fn insert(&mut self, key: Key, node: LeafNode, next_separator: Option<Key>) {
-        let entry = self.inner.entry(key).or_insert_with(|| ChangedLeafEntry {
-            deleted: None,
-            inserted: None,
-            next_separator: None,
-        });
+    pub fn insert_leaf(&mut self, key: Key, node: LeafNode, next_separator: Option<Key>) {
+        if let Some(last_entry_next_separator) = self.get_last_next_separator() {
```
rphmeier, line 80: naming: `last_next_separator` sounds like an oxymoron. any ideas for a clearer term?

nomt/src/beatree/ops/update/leaf_stage.rs
```diff
@@ -35,50 +37,95 @@ pub struct ChangedLeafEntry {
...
+            // there is a range of unchanged leaves
+            if last_entry_next_separator < key {
+                self.insert_range(last_entry_next_separator, Some(key));
```
rphmeier, line 83: It seems like this computation is something that the receiver can also just perform, so it'd save on some data and LoC to do it there

nomt/src/beatree/ops/update/leaf_stage.rs
```diff
@@ -285,42 +333,85 @@ fn try_answer_left_neighbor(
...
+        .map(|_| {
+            leaves_tracker
+                .inner
+                .pop_first()
+                .and_then(|(key, entry)| match entry {
+                    // SAFETY: the existance and variant of `take` leaves_tracker items was previously checked
+                    LeavesTrackerEntry::ChangedLeaf(changed_entry) => Some((key, changed_entry)),
+                    _ => None,
+                })
+                .unwrap()
```
rphmeier, line 381: why does this `unwrap` still hold?

rphmeier (approved):
Looks broadly OK, but I think it could be conceptually simplified (in follow ups) as mentioned in comments

nomt/src/beatree/ops/update/leaf_stage.rs
gabriele-0201: Yes, initially, I didn't have a new_base field, but it was possibly inside the last item of `changed`. I moved it to `deleted` and `new_base` just to create less messy code when dealing with `LeavesTrackerEntry`. I'm not a fan of that enum, and as you said in another comment, we could get rid of it by extracting the range from the changed items.

Therefore, to avoid refactoring everything in this PR, I will keep this now and work on a simplification follow-up!

nomt/src/beatree/ops/update/leaf_stage.rs
gabriele-0201: I think that removing it would make things less obvious but with cleaner code. I will add more comments to compensate for the absence of something explicit (in a follow-up)

nomt/src/beatree/ops/update/leaf_stage.rs
gabriele-0201: yes! avoid saving the `UnchangedRange` in the tracker and let the receiver extract the information seems to be the  way to go!

nomt/src/beatree/ops/update/leaf_stage.rs
gabriele-0201: my bad, during some refactoring I forgot to move and update a comment a little bit above, instead of `SAFETY` it should be `UNWRAP` and it explains why the `unwrap` holds

nomt/src/beatree/ops/update/leaf_stage.rs
gabriele-0201: yes, it could for sure be better named. I could go for `higher_next_separator`

rphmeier:
### Merge activity

* **Oct 8, 3:02 PM EDT**: @rphmeier started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/426).

PR #428, opened 2024-10-04
feat: introduce seglog

pepyakin:
This commit introduces a segmented log (seglog for short). It is an
append-only log that allows pruning old items.

The seglog module provides a storage abstraction that supports appending
data sequentially. It is designed to handle large volumes of data
efficiently by segmenting the log into manageable parts. Key features
include:

- **Append-Only**: Data can only be appended, ensuring a consistent and
  ordered log.
- **Pruning**: Old data can be pruned from the beginning of the log to
  reclaim space, keeping the log size manageable.
- **Segment Management**: The log is divided into segments, each with a
  maximum size limit. New segments are created as needed, and old
  segments can be removed once they are no longer needed.
- **Record Management**: Each record has a unique ID, and the module
  ensures that records are written and read in order.
- **Recovery**: The module supports recovery from unclean shutdowns,
  ensuring data integrity and consistency.

### DIFF
--- a/nomt/src/seglog/mod.rs
+++ b/nomt/src/seglog/mod.rs
@@ -1,0 +1,1344 @@
+//! A segment log is a file that provides a storage abstraction which allows appending data.
+//!
+//! The segment log is not suitable for random access. The only way to read data from the
+//! segment log is to start from the oldest record and read all the way to the newest upon recovery.
+//!
+//! The segment log allows pruning of old data (from the beginning of the file) to reclaim space.
+//!
+//! The log is designed to operate efficiently with a small number of segments. It is the
+//! responsibility of the user to prune the log periodically to maintain this efficiency.
+//!
+//! `max_segment_size` is a crucial parameter that determines the maximum size of a segment.
+//! When setting this parameter, users should consider several factors:
+//! 1. It should be large enough to prevent the creation of too many segments for the anticipated workload.
+//! 2. It should not be excessively large, as this would make it harder to reclaim space efficiently.
+//! 3. It should be balanced to allow for efficient pruning, as pruning involves linearly scanning the file.
+//!
+//! The user must carefully choose this parameter based on their expected record sizes and
+//! append/prune patterns. The log is pruned on a segment basis, and a segment cannot be
+//! removed if it still contains any used elements.
+//!
+//! There is a limit on the maximum size of a record payload, which is currently set to 1 GiB.
+//! However, this limit may be subject to change in future versions.
+
+use anyhow::{ensure, Context, Result};
+use std::{
+    fmt,
+    fs::{self, File, OpenOptions},
+    io::{Seek, SeekFrom},
+    mem,
+    path::{Path, PathBuf},
+};
+
+mod segment_filename;
+mod segment_rw;
+
+use self::segment_rw::{SegmentFileReader, SegmentFileWriter};
+
+const RECORD_ALIGNMENT: u32 = 4096; // 4K alignment
+const HEADER_SIZE: u32 = 12; // 8 bytes for record ID, 4 bytes for payload length
+const MAX_RECORD_PAYLOAD_SIZE: u32 = 1 << 30; // 1 GiB
+
+/// A record ID.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct RecordId(pub u64);
+
+impl RecordId {
+    pub fn nil() -> Self {
+        RecordId(0)
+    }
+
+    pub fn from_bytes(bytes: [u8; 8]) -> Self {
+        RecordId(u64::from_le_bytes(bytes))
+    }
+
+    pub fn next(&self) -> Self {
+        RecordId(self.0 + 1)
+    }
+
+    pub fn prev(&self) -> Option<RecordId> {
+        if self.0 == 0 {
+            None
+        } else {
+            Some(RecordId(self.0 - 1))
+        }
+    }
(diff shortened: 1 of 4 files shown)


pepyakin:
* **#428** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/428?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

pepyakin:
v2 introduces pruning of the latest items. I missed that during the rewrite and discovered when porting the rollback to use seglog. This is needed upon the rollback: it undoes some latest commits and they should be discarded from the log to free up space for the new commits for the same height.

nomt/src/seglog/mod.rs
```diff
@@ -0,0 +1,1267 @@
...
+mod segment_filename;
+mod segment_rw;
+
+use self::segment_rw::{SegmentFileReader, SegmentFileWriter};
+
+const RECORD_ALIGNMENT: u32 = 4096; // 4K alignment
+const HEADER_SIZE: u32 = 12; // 8 bytes for record ID, 4 bytes for payload length
+const MAX_RECORD_PAYLOAD_SIZE: u32 = 1 << 30; // 1 GiB
+
+// TODO: rename active to live.
```
rphmeier, line 43: for a follow-up, I guess?

nomt/src/seglog/segment_rw.rs
```diff
@@ -0,0 +1,462 @@
...
+    /// Must be called after `read_header`.
+    pub fn read_payload(&mut self, payload_buf: &mut Vec<u8>) -> Result<()> {
+        payload_buf.resize(self.payload_length.unwrap() as usize, 0);
+        self.buf_reader.read_exact(payload_buf)?;
+        self.seek_next()?;
+        Ok(())
+    }
+
+    /// Positions the reader at the beginning of the next record.
+    pub fn seek_next(&mut self) -> Result<()> {
```
rphmeier, line 153: worth adding a comment (here and elsewhere) that usage panics when the header hasn't yet been read.

nomt/src/seglog/segment_filename.rs
```diff
@@ -0,0 +1,120 @@
+use anyhow::{Context, Result};
+
+pub fn format(prefix: &str, segment_id: u32) -> String {
+    // The format string specifies a 10-digit number, so we pad with leading zeros from
+    // the left. This assumes that segment_id is a 32-bit integer, which is confirmed by
```
rphmeier, line 5: can you humor me with a back-of-envelope calculation about whether 32 bits is enough?

nomt/src/seglog/mod.rs
```diff
@@ -0,0 +1,1267 @@
...
+        }
+        self.active_segment_start.is_some() && self.active_segment_end.is_none()
+    }
+
+    // /// Find the index of the first and last active segments.
+    // ///
+    // /// An active segment is a segment that contains records in the active range.
+    // ///
+    // /// Also makes some checks on the segments.
+    // fn find_active_segments(&mut self) -> Result<RangeInclusive<usize>> {
```
rphmeier, line 549: is this meant to be committed? will it be reinstated in a follow up?

pepyakin: Meant to be removed.

nomt/src/seglog/segment_filename.rs
pepyakin: Assume you create a segment for each commit. Personally, I am assuming arguably a conservative case where we do 3 commits per second. If so, it would take around 10 years (2^32 s / 3) of such usage to exhaust the whole range. That, together with the another assumption that we assume frequent resyncs, it looks fine.

Moreover, to actually trigger a problem, you need start overwriting the data, which can only happen without pruning, but with any reasonable segment size you would exhaust modern disks first before you get to that point. 

There is still a slight problem though: we should use wrapping_add for segment_id addition, and in fact, in the original rollback PR it did use wrapping_add, but it was lost after this rewrite. It should be fine without it in practice because without debug_assertions on it would still wrap around but ideally we still file a fix.

pepyakin: Actually, now thinking about it, it would be a tad more complicated than using wrapping_add. This is because we rely on IDs to imply ordering upon recovery. So if a recovery does happen when there are segment ID values around the both sides of u32::MAX, then the segments would be presented in a wrong order and the recovery would fail because it would detect the mismatch in record IDs.

I would still say that 32-bit is ought to be enough for everyone and this would be fine for now.

nomt/src/seglog/mod.rs
pepyakin: did it in this. Check v3.

nomt/src/seglog/mod.rs
pepyakin: Check v3.

nomt/src/seglog/segment_rw.rs
pepyakin: check v3

pepyakin (commented):
Submitted v3 with the suggested changes.

pepyakin:
v4, v5 adds `RecordId::prev` and some fixes which are required for the upstream rollback PR. It also adds some tests and docs for those changes.

nomt/src/seglog/segment_filename.rs
rphmeier: Yeah, this is convincing enough, but I wonder - why not use a u64. is there a downside to it?

PR #447, opened 2024-10-10
feat: remove tentative_*_slot APIs, replace with read and warm-up

rphmeier:
This PR removes the `tentative_read_slot` and `tentative_write_slot` APIs from `Session` and replaces them with `read(key) -> value` and `warm_up(key)` APIs.

There are motivations for this change:
  1. Supporting warm-ups with app-level caching. If the user has an app-level value cache, they need some way to tell NOMT to warm up the path. This requires a `warm_up` function to exist.
  2. Avoiding redundant warm-ups with read-then-write patterns. In the current state of the API, we call `warm_up` twice for a key: once when we read it, and once when we write it.
    - This inefficiency might not be avoidable in general computation, but at least this way the user has the ability to elide redundant calls as far as their execution environment permits.
  3. Avoiding warm-ups for reads when not merkle-proving. If the user isn't generating merkle proofs, then only writes need to be warmed up. The current state of the API doesn't account for this.

### DIFF
--- a/benchtop/src/nomt.rs
+++ b/benchtop/src/nomt.rs
@@ -132,7 +132,9 @@ impl<'a> Transaction for Tx<'a> {
         match self.access.entry(key_path) {
             Entry::Occupied(o) => o.get().last_value().map(|v| v.to_vec()),
             Entry::Vacant(v) => {
-                let value = self.session.tentative_read_slot(key_path).unwrap();
+                let value = self.session.read(key_path).unwrap();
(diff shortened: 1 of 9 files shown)


rphmeier:
* **#447** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/447?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

PR #450, opened 2024-10-10
test: add tests for leaf updater

rphmeier:

PR #451, opened 2024-10-10
fix: left-shift bug in reconstruct_key

rphmeier:
I was running into an issue where the separator was being shifted left by 5 bits but then only getting 3 bits from the next byte.

e.g. `11111111` was becoming `11100000` and then `11100111`, which was wrong. The initial fix just changed an `amount` to `(8 - amount)`.

But the `mask` didn't seem useful in this case (it was being generated as `11101111`) so I removed it. What was it for? Is it still necessary?

### DIFF
--- a/nomt/src/beatree/ops/bit_ops.rs
+++ b/nomt/src/beatree/ops/bit_ops.rs
@@ -115,11 +115,10 @@ pub fn reconstruct_key(maybe_prefix: Option<RawPrefix>, separator: RawSeparator)
 
         let mut chunk_shifted = chunk.to_be_bytes();
 
-        // move bits remainder between chunk bounderies
+        // move bits remainder between chunk boundaries
         match &mut shift {
             Some(Shift::Left(amount)) if chunk_index < n_chunks - 1 => {
-                let mask = !(1 << (8 - *amount) - 1);
-                let remainder_bits = (separator_bytes[(chunk_index + 1) * 8] & mask) >> *amount;
+                let remainder_bits = (separator_bytes[(chunk_index + 1) * 8]) >> (8 - *amount);
                 chunk_shifted[7] |= remainder_bits;
             }
             Some(Shift::Right(_, prev_remainder, curr_remainder)) => {


rphmeier:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/451?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#452** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/452?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#451** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/451?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#450** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/450?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

gabriele-0201:
Interesting, it didn't cause many issues and tests were still passing (that's why more restrictive unit tests are needed for this bit ops)

Anyway, I think that's just a mistake made by me. Shifting right by `amount` is clearly wrong. As you said, the fix was to shift it by `8 - amount` because we only want the most significant `amount` bits. I think I kept working on that mistake and probably thought that masking bits was needed before shifting. However, that's not true if we only want the significant bits. The unnecessary bits are removed by the shift itself, and all other bits will be filled with zeros

gabriele-0201 (approved):

rphmeier:
> The unnecessary bits are removed by the shift itself, and all other bits will be filled with zeros

Yep! Thanks for clarifying that the mask was mostly a workaround. IMO this issue is a result of lack of tests on `reconstruct_key`, because it should have been caught if we were testing all combinations of start/end bit indices.

rphmeier:
### Merge activity

* **Oct 11, 11:28 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/451).

PR #469, opened 2024-10-14
tests: minimal reproducible example

pepyakin:


pepyakin:
* **#469** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/469?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

pepyakin:
I've tried to bisect this issue, however, I couldn't find the "good" commit even though I descended quite deep into the history.

rphmeier:
As far as I can see, this is fixed after #474

The bug isn't in page-walker per se, but seek.

Do you want these merged in-tree? They need some modifications if so:
  1. naming tests `test_` is bad style
  2. it's better to delete the existing DB, run the test, and then run it again (without deleting) to avoid DB format issues from breaking tests
  
  
Also, I'm not sure if these are _really_ a minimal example of something you encountered. I am having a hard time understanding how you encountered keys with 255 shared bits as part of an innocent testing of rollback. Can you explain?

pepyakin:
I don't think this should be merged. I haven't looked yet, but I bet G's test would be more relevant. I also agree with the suggestions.

Those were just demonstrations. Those are minimal relative to the full package coming in the tests of [rollback PR](https://github.com/thrumdev/nomt/blob/73a22018352ca646c8838b6a46f64cec4f1892d4/nomt/tests/rollback.rs). I didn't really look deep into it because I didn't really want to get distracted from the rollback PR, so really appreciated that @gabriele-0201 took a look. IOW, I don't really know what happened and if all the issues were caught by it.

Some of the problems got fixed by cleaning up the directory. I don't really know if we just sweeped the problem under a rug and there is an issue on the persist or recovery paths.

Closing this for now.

PR #473, opened 2024-10-15
feat: rollback

pepyakin:
This changeset implements a rollback feature.

The rollback feature allows NOMT to roll back to
a previous commit state.

It is implemented by storing a reverse delta for
each commit. When rolling back, we apply the reverse
delta to the current state to restore the previous
state.

To gather the reverse delta NOMT collects the
prior values of the keys that are being modified
or removed. When committing, it uses this information
to create a reverse delta. To that end,
the Session object has a `preserve_prior_value` method
that the caller should call for each key that they wish
to preserve.

The definitive set of values to be committed is
determined by the `Nomt::commit` call. As such,
there is no issue with invoking this function for
values that may not ultimately be written.
Similarly, values that were not marked for
preservation during the session but are included
in the final set will still be preserved. However,
if the user does not call `preserve_prior_value`
then the prior values will be fetched during the
commit process which degrades performance.

### DIFF
--- a/benchtop/src/nomt.rs
+++ b/benchtop/src/nomt.rs
@@ -169,5 +169,6 @@ impl<'a> Transaction for Tx<'a> {
         }
 
         self.session.warm_up(key_path);
+        self.session.preserve_prior_value(key_path);
     }
 }
--- a/nomt/src/lib.rs
+++ b/nomt/src/lib.rs
@@ -42,6 +42,7 @@ mod page_cache;
 mod page_diff;
 mod page_region;
 mod page_walker;
+mod rollback;
 mod rw_pass_cell;
 mod seek;
 mod seglog;
@@ -123,6 +124,14 @@ impl KeyReadWrite {
                 v.as_deref()
             }
         }
+    }
+
+    /// Returns true if the key was written to.
+    pub fn is_write(&self) -> bool {
+        matches!(
+            self,
+            KeyReadWrite::Write(_) | KeyReadWrite::ReadThenWrite(_, _)
+        )
     }
 
     /// Updates the state of the given slot.
@@ -227,6 +236,16 @@ impl Nomt {
         self.root() == TERMINATOR
     }
 
+    /// Returns the value stored under the given key.
+    ///
+    /// Returns `None` if the value is not stored under the given key. Fails only if I/O fails.
+    ///
+    /// This is used for testing for now.
+    #[doc(hidden)]
+    pub fn read(&self, path: KeyPath) -> anyhow::Result<Option<Value>> {
+        self.store.load_value(path)
+    }
+
     /// Creates a new [`Session`] object, that serves a purpose of capturing the reads and writes
     /// performed by the application, updating the trie and creating a [`Witness`], allowing to
     /// re-execute the same operations without having access to the full trie.
@@ -234,12 +253,22 @@ impl Nomt {
     /// Only a single session may be created at a time. Creating a new session without dropping or
     /// committing an existing open session will lead to a panic.
     pub fn begin_session(&self) -> Session {
+        self.begin_session_inner(/* allow_rollback */ true)
+    }
+
+    fn begin_session_inner(&self, allow_rollback: bool) -> Session {
         let prev = self
             .session_cnt
             .swap(1, std::sync::atomic::Ordering::Relaxed);
         assert_eq!(prev, 0, "only one session could be active at a time");
+        let store = self.store.clone();
+        let rollback_delta = if allow_rollback {
+            self.store.rollback().map(|r| r.delta_builder())
+        } else {
+            None
+        };
         Session {
-            store: self.store.clone(),
+            store,
             committer: Some(self.commit_pool.begin(
                 self.page_cache.clone(),
                 self.page_pool.clone(),
@@ -248,6 +277,7 @@ impl Nomt {
             )),
             session_cnt: self.session_cnt.clone(),
             metrics: self.metrics.clone(),
+            rollback_delta,
         }
     }
 
@@ -304,6 +334,11 @@ impl Nomt {
                 );
             }
         }
+        if let Some(delta_builder) = session.rollback_delta.take() {
+            // UNWRAP: if rollback_delta is `Some``, then rollback must be also `Some`.
+            let rollback = self.store.rollback().unwrap();
+            rollback.commit(self.store.clone(), &actuals, delta_builder)?;
+        }
 
         let mut compact_actuals = Vec::with_capacity(actuals.len());
         for (path, read_write) in &actuals {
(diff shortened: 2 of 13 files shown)


benchtop/src/nomt.rs
```diff
@@ -169,5 +169,6 @@ impl<'a> Transaction for Tx<'a> {
         }
 
         self.session.warm_up(key_path);
+        self.session.preserve_prior_value(key_path);
```
rphmeier, line 172: just a note (as an issue that existed already)

if we write the same slot twice, these calls are redundant. but our current workloads typically won't.

nomt/src/lib.rs
```diff
@@ -227,19 +236,39 @@ impl Nomt {
...
+    }
+
+    fn begin_session_inner(&self, allow_rollback: bool) -> Session {
         let prev = self
             .session_cnt
             .swap(1, std::sync::atomic::Ordering::Relaxed);
         assert_eq!(prev, 0, "only one session could be active at a time");
+        let store = self.store.clone();
+        let rollback_delta = if allow_rollback {
+            self.store.rollback().map(|r| r.detla_builder())
```
rphmeier, line 266: typo: should be delta_builder?

nomt/src/lib.rs
```diff
@@ -335,6 +370,38 @@ impl Nomt {
         Ok((new_root, commit.witness, commit.witnessed_operations))
     }
 
+    /// Perform a rollback of the last `n` commits.
+    ///
+    /// This function assumes no other sessions are active.
```
rphmeier, line 375: The documentation is somewhat unclear here. In particular the meaning of 'no other sessions' is unclear, since this function is on `Nomt`, not `Session`.

Should it say "This function assumes no sessions are active"?

nomt/src/store/mod.rs
```diff
@@ -54,6 +58,20 @@ impl Store {
             create(o)?;
         }
 
+        // TODO: fix TOCTOU race and lock the directory.
```
rphmeier, line 61: this should be a filed issue!

nomt/src/rollback/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+#[cfg(test)]
+mod tests;
+
+const MAX_SEGMENT_SIZE: u64 = 64 * 1024 * 1024; // 64 MiB
+
+struct InMemory {
+    /// The log of deltas that we have accumulated so far modulo the [`Self::head`].
+    ///
+    /// The items are pushed onto the back and popped from the front. When the log reaches
+    /// `limit_log_len`, the oldest delta is discarded.
```
rphmeier, line 40: `limit_log_len` isn't defined

nomt/src/rollback/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+            db_dir_path,
+            db_dir_fd,
+            "rollback".to_string(),
+            MAX_SEGMENT_SIZE,
+            rollback_start_active.into(),
+            rollback_end_active.into(),
+            |record_id, payload| {
+                let mut cursor = Cursor::new(payload);
+                let delta = Delta::decode(&mut cursor)?;
+                in_memory.push_back(record_id, delta);
```
rphmeier, line 108: is this loading the entire rollback log into memory, or just the head?

AFAIK one requirement of the feature is that the rollback log may be very large. For example, in Geth 1.13.0 they keep 90k blocks in their rollback log.

nomt/src/rollback/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+                Ok(())
+            },
+        )?;
+        let shared = Arc::new(Shared {
+            worker_tp: ThreadPool::new(4),
```
rphmeier, line 113: should this be parameterized somehow in `Options`?

rphmeier: It's not clear this needs to be a thread-pool, either - one thread is usually sufficient for saturating I/O. But I guess that'd require an asynchronous version of `load_value`.

rphmeier: Let's file a follow-up for that

nomt/src/rollback/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+                if let Some((path, value)) = tentative_priors.remove(path) {
+                    // Tentative speculation was a hit. Keep the entry.
+                    final_priors.insert(path, value);
+                } else {
+                    // The delta builder was not aware of this write. Initiate a fetch from the store
+                    // and record the result as a prior.
+                    let store = store.clone();
+                    let path = path.clone();
+                    let final_priors = final_priors.clone();
+                    self.tp.execute(move || {
```
rphmeier, line 326: This can be avoided if `read_write` is a `ReadThenWrite`

nomt/src/rollback/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+}
+
+impl ReverseDeltaBuilder {
+    /// Note that a write might be made to a key and that the rollback should preserve the prior
+    /// value. This function is speculative; the rollback delta may later be committed with a
+    /// different set of operations, and some of the tentative operations may be discarded.
+    ///
+    /// This function doesn't block.
+    pub fn tentative_preserve_prior(&self, store: impl LoadValue, key_path: KeyPath) {
```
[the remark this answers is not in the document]
rphmeier: Well, considering my comment on L326, it seems that this function only needs to be called for 'bare' writes.

rphmeier (commented):
This looks broadly good, but I have a couple of concerns, none of them blocking. But we should discuss how to address them for the final product.

1. The thread pool for populating the delta seems unnecessary. We most likely need only a single thread to saturate the I/O queue.
2. As implemented, this only supports short rollbacks, not long ones, since all deltas are also stored in memory. My understanding was that this feature was to be loosely based off of the work in https://blog.ethereum.org/2023/09/12/geth-v1-13-0 .

pepyakin (commented):
Alright, submitted v9 with the fixes. Specifically:

- fix the typo of delta_builder
- fix the outdated docs in rollback/mod.rs
- updated the instructions on how to use the preserve_prior_value
- updated the doc on `Nomt::rollback`

I agree with the other comments and we should leave for follow-ups:

- create an issue for TOCTOU. Just a note that this is an existing issue and I just spelled it out. Also, this is more of a safeguard, since the assumptions of the database is that it has a complete ownership of the directory.
- allow the rollback to not be completely loaded into the memory. This is considerable rework, but doable.
- reduce the number of threads or make it configurable. This requires testing.
- avoid preserving writes if there is a read for that value in actuals and document that it's not necessary to issue `preserve_prior_value` if that value is being read.

PR #474, opened 2024-10-15
fix: stops seeker at last layer of the trie

gabriele-0201:
This prevents the seeker from trying to construct an invalid `page_id` calling `next_page_id` on a `page_id` part of the last layer of the trie, which resulted in a panic in `core/src/trie_pos.rs:196`

PR #480, opened 2024-10-16
test: add last_layer_trie test

gabriele-0201:
test change introduced in #474

### DIFF
--- a/nomt/tests/last_layer_trie.rs
+++ b/nomt/tests/last_layer_trie.rs
@@ -1,0 +1,39 @@
+mod common;
+
+use common::Test;
+
+#[test]
+fn last_layer_trie() {
+    let mut t = Test::new_with_params(
+        "last_layer_trie", // name
+        1,                 // commit_concurrency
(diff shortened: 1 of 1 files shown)


gabriele-0201:
* **#480** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/480?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

pepyakin:
1. If we wanted a proper randomness, why not go for `rand::thread_rng()` as everywhere else in the codebase?
2. I am not sure if we want to have a non-deterministic test in the first place since those lead to flaky tests.
3. just to clarify, have you confirmed that this test fails without #474 applied?

gabriele-0201:
1/2. Upon rethinking about it, using randomness here only introduces problems, as you mentioned, leading to flaky tests
3. Yes! 

I will update with some hardcoded keys!

pepyakin (approved):

PR #491, opened 2024-10-18
opt(rollback): don't fetch read-then-write

pepyakin:
This changeset implements an optimization suggested by @rphmeier to
elide fetches to the store for the keys that were passed in Nomt::commit
as ReadThenWrite.

Note that this doesn't check if that was the actual value and blindly
trusts the user to correctly specify the value, as doing so will
actually makes no sense performance-wise.


pepyakin:
* **#491** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/491?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

rphmeier (approved):

PR #492, opened 2024-10-18
fix: stop split from creating too big right leaf

gabriele-0201:
Another solution could be to find another `LEAF_BULK_SPLIT_THRESHOLD` but I couldn't find any value which was passing the tests introduced in #493, thus a recursive call to split was the easiest solution to cover this edge case

### DIFF
--- a/nomt/src/beatree/ops/update/leaf_updater.rs
+++ b/nomt/src/beatree/ops/update/leaf_updater.rs
@@ -324,7 +324,13 @@ impl LeafUpdater {
             right_gauge.ingest(item_size);
         }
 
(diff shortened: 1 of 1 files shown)


gabriele-0201:
* **#492** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/492?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

rphmeier (approved):
Nice catch!

Doesn't branch have the same problem?

gabriele-0201:
Yes, the branch updater should have the same problem. I was just focused on the leaf stage, so I pushed the fix and tests only for this. I will, of course, push it also for the branch in a following pr!

PR #498, opened 2024-10-21
feat: configurable rollback tp size

pepyakin:
We discussed potentially changing the way we dispatch the loads to async IO which would obsolete this setting, but I decided to submit this anyway in case for the time being.

PR #500, opened 2024-10-21
feat: flock db dir

### LINKED ISSUE #483: Obtain exclusive lock for db dir
We build upon the assumption that NOMT owns the directory exclusively. Right now this is not enforced in anyway and indeed this sometimes shows in tests when they run in parallel and overwrite each others data. We should probably look into flock or something to reject opening the directory and the files inside it by multiple processes.

pepyakin:
Closes #483

This commit aim is to return an error message if another process tries
to use the same db directory, which is unsupported according to our
assumptions.

Note that this uses only advisory locking. It's still possible to mess
with the directory and containing files. It's the responsibility of the
user to prevent that and failure to do so can and will lead to
unexpected behavior.

## Implementation Notes

A couple of words on my research of the problem and the taken approach.

There are mandatory and advisory locking. Mandatory locking is a feature of OS that reserves an exclusive right for a specific program to access a particular file. On Linux, mandatory locking is turned off by default and to be avoided (see [this](https://www.kernel.org/doc/Documentation/filesystems/mandatory-locking.txt)). macOS doesn't support mandatory locking.

The typical way to solve this issue is to use advisory locking. That is, depend on the cooperation between the processes that use the files and directories in question. Obviously, this depends on a good will of those processes and not a ironclad solution.

On Linux, it's possible to lock a directory. `flock` with the fd of the directory open with O_PATH can be used. macOS doesn't support opening fds with O_PATH, so this option is not available. To make the solution cross-platform, I stuck to creating a lock file named `.lock` in the DB directory and obtaining an exclusive lock on that.

Now, the original TODO and issue (#483) mentions the TOCTOU issues. Namely, the fact that the checking if the db directory exists, opening the db dir, initializing the db, opening the db files are all non-atomic and it's possible that the state of the filesystem has changed between the checks and file opening or creation, which can lead to all sorts of issues. 

Proper implementation should probably employ `openat` family of syscalls to avoid some of those, at least on Linux. This all is out-of-scope for this PR and better be addressed on the next big refactoring in one fell swoop, e.g. with #501.

### DIFF
--- a/nomt/Cargo.toml
+++ b/nomt/Cargo.toml
@@ -29,6 +29,7 @@ lru = "0.12.3"
 libc = "0.2.155"
 criterion = { version = "0.3", optional = true }
 thread_local = "1.1.8"
+fs2 = "0.4.3"
 
 [target.'cfg(target_os="linux")'.dependencies]
 io-uring = "0.6.4"
--- a/nomt/src/store/flock.rs
+++ b/nomt/src/store/flock.rs
@@ -1,0 +1,41 @@
+//! This module provides a cross-platform advisory lock on a directory.
+
+use std::{
+    fs::{File, OpenOptions},
+    path::Path,
+};
+
+use fs2::FileExt as _;
+
+/// Represents a cross-platform advisory lock on a directory.
+pub struct Flock {
+    lock_fd: File,
+}
+
+impl Flock {
+    pub fn lock(db_dir: &Path, lock_filename: &str) -> anyhow::Result<Self> {
+        let lock_path = db_dir.join(lock_filename);
+
+        let lock_fd = OpenOptions::new()
+            .read(true)
+            .write(true)
+            .create(true)
+            .open(lock_path)?;
+
+        match lock_fd.try_lock_exclusive() {
+            Ok(_) => Ok(Self { lock_fd }),
+            Err(_) => {
+                let err = fs2::lock_contended_error();
+                anyhow::bail!("Failed to lock directory: {err}");
+            }
+        }
+    }
+}
+
+impl Drop for Flock {
+    fn drop(&mut self) {
+        if let Err(e) = self.lock_fd.unlock() {
+            eprintln!("Failed to unlock directory lock: {e}");
+        }
+    }
+}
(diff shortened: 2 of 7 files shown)


pepyakin:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/500?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#500** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/500?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#498** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/498?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

nomt/Cargo.toml
```diff
@@ -29,6 +29,7 @@ lru = "0.12.3"
 libc = "0.2.155"
 criterion = { version = "0.3", optional = true }
 thread_local = "1.1.8"
+fs2 = "0.4.3"
```
pepyakin, line 32: This is not my favorite, as this doesn't give us much. We don't care about supporting much besides modern Linux and macOS, we don't explicitly target Windows and can just directly use `flock`. But I thought for now this is OK and we should reconsider that in #501.

nomt/src/store/flock.rs
```diff
@@ -0,0 +1,41 @@
...
+                anyhow::bail!("Failed to lock directory: {err}");
+            }
+        }
+    }
+}
+
+impl Drop for Flock {
+    fn drop(&mut self) {
+        if let Err(e) = self.lock_fd.unlock() {
+            eprintln!("Failed to unlock directory lock: {e}");
```
pepyakin, line 38: I know this is the first appearance of `eprintln!` in this codebase. Should we finally add `tracing` or `log`? Or should we just swallow this error? Or leave it as is for now?

rphmeier (approved):

pepyakin:
### Merge activity

* **Oct 21, 12:26 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/500).

PR #554, opened 2024-11-05
fix: don't use O_DIRECT on tmpfs

pepyakin:
We create a tempdir and open the file with the O_DIRECT flag.
Surprisingly, nobody ever noticed that, but it shouldn't succeed 
because tmpfs doesn't support O_DIRECT.


rphmeier (approved):

rphmeier:
### Merge activity

* **Nov 5, 4:25 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/554).

PR #566, opened 2024-11-18
chore:add,update prints in leaf/branch_stage tests

gabriele-0201:
`println`s aims to simplify debugging in case of a bug found in the update logic  ~A future pr will contain a refactor of those prints to be part of a dump if the test fails.~

edit. it's not worth adding complexity to print things on CI if the test is easily reproducible locally


rphmeier (commented):
I don't like having printlns committed, even in tests. Often, the prints that are best to add are very dependent on the issue you're debugging. I would prefer to reject this change, even if it means adding the prints redundantly when debugging.

rphmeier (commented):
The exception being when the test is random, and we need to somehow save the seed. But that should be it!

PR #567, opened 2024-11-18
refactor: add bitwise_memcpy

gabriele-0201:
This pr not only extracts the shift and copy bits logic a new function(`bitwise_mempcy`), but also includes a slight update: now `bitwise_memcpy` leaves unchanged all the bits that are not part of the copy in the destination

edit. the pr now also includes a replacement of `init` with `start` every time it was not used correctly

PR #565, opened 2024-11-18
refactor: follow-up to leaf updater uses binary search

gabriele-0201:
The following pr contains all the follow-ups to #529.

One specific thing I would like to ask is an opinion to the introduction of `LeafNode::extract_key`. It is a small abstraction that still requires some conceptual thoughts. The point is: the layout of the leaf node is mostly encapsulated within its module, and inside `keep_up_to`, I was extracting the key from the `cell_pointers`. However, I thought that a modification in the layout of the leaf node could silently break the leaf updater. That's why I added this helper function. In case a layout change happens, the entire `leaf/node.rs` code would be updated, including the way a key is extracted from the cell pointer.

PR #564, opened 2024-11-18
fix: update BaseLeaf when keep_up_to jumps to end

gabriele-0201:

PR #568, opened 2024-11-18
[WIP] opt: add branch builder push_chunk

gabriele-0201:
I'm not a fan of having introduced `RawSeparators`, which include the old `RawSeparator`, but having both seems very redundant.
If you have other ideas, they are welcome!

### DIFF
--- a/nomt/src/beatree/branch/node.rs
+++ b/nomt/src/beatree/branch/node.rs
@@ -1,8 +1,15 @@
+use std::ops::Range;
+
 use bitvec::prelude::*;
 
 use super::BRANCH_NODE_SIZE;
-use crate::beatree::Key;
-use crate::io::{FatPage, PagePool};
+use crate::{
+    beatree::{
+        ops::{bit_ops::bitwise_memcpy, get_key},
+        Key,
+    },
+    io::{FatPage, PagePool},
+};
 
 // Here is the layout of a branch node:
 //
@@ -107,8 +114,29 @@ impl BranchNode {
         self.as_mut_slice()[start..].view_bits_mut()[..prefix_len].copy_from_bitslice(prefix);
     }
 
-    pub fn raw_separator(&self, i: usize) -> RawSeparator {
+    fn cells_mut(&self) -> &mut [[u8; 2]] {
+        unsafe {
+            std::slice::from_raw_parts_mut(
+                self.page[BRANCH_NODE_HEADER_SIZE..BRANCH_NODE_HEADER_SIZE + 2].as_ptr()
+                    as *mut [u8; 2],
+                self.n().into(),
+            )
+        }
+    }
+
+    pub fn raw_separator(&self, i: usize) -> RawSeparators {
         self.view().raw_separator(i)
+    }
+
+    pub fn raw_separators_mut(&mut self, from: usize, to: usize) -> RawSeparatorsMut {
+        let RawSeparatorsData {
+            start,
+            byte_len,
+            bit_init,
+            bit_len,
+        } = self.view().raw_separators_data(from, to);
+
+        (&mut self.page[start..start + byte_len], bit_init, bit_len)
     }
 
     fn set_separator(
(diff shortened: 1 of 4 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/568?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#568** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/568?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/568?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#567** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/567?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#566** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/566?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#565** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/565?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#564** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/564?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

gabriele-0201:
During the optimization of the branch_stage, I have just found out that it will likely be easier and more accurate to avoid having chunks intersect with non-compressed items. There are multiple reasons for this, and the decision is still pending, heavily dependent on the code I'm working on to introduce binary_search into the branch_stage. Therefore, I will move back this PR to WIP

nomt/src/beatree/branch/node.rs
```diff
@@ -107,10 +114,31 @@ impl BranchNode {
         self.as_mut_slice()[start..].view_bits_mut()[..prefix_len].copy_from_bitslice(prefix);
     }
 
-    pub fn raw_separator(&self, i: usize) -> RawSeparator {
+    fn cells_mut(&self) -> &mut [[u8; 2]] {
```
rphmeier, line 117: this needs to be `&mut`

nomt/src/beatree/branch/node.rs
```diff
@@ -217,18 +285,43 @@ impl<'a> BranchNodeView<'a> {
...
+    bit_len: usize,
+}
+
+// A RawSeparators is made by a triple, the raw bytes, the bit-offset
+// at which the separators starts to be encoded in the first byte
 // and the relative bit length
 //
 // The raw bytes are always a multiple of 8 bytes in length
-pub type RawSeparator<'a> = (&'a [u8], usize, usize);
+pub type RawSeparators<'a> = (&'a [u8], usize, usize);
```
rphmeier, line 323: it's odd to define this type as a plural but then use it like a singular in e.g. `fn raw_separator` or `fn reconstruct_key`

nomt/src/beatree/branch/node.rs
```diff
@@ -294,11 +387,305 @@ impl BranchNodeBuilder {
         self.index += 1;
     }
 
+    pub fn push_chunk(&mut self, base: &BranchNode, from: usize, to: usize) {
```
rphmeier, line 390: Looking at this function, it looks to me like it internalizes too much complexity, especially around handling edge cases.

the contract of this function is not clear. the typical case I expect we would optimize for is basically just a single bitwise memcpy of the separators and a normal memcpy of the node pointers, which is what we get when the prefix hasn't changed.

I don't see a strong case that this would be more efficient than calling `push` repeatedly in the edge cases, and we've also pessimized the base case with a bunch of branches and additional logic.

It seems better to have alternative variants for each expected code path.

My suggestion:
  - Have `push_chunk` variants for all changes in the prefix (none, shrink, grow). Our base case is none.
  - Have `push_chunk` variant for uncompressed entries.
  - Have the branch updater figure out which one to call in which situations. Figuring out which to call seems pretty easy during `BranchUpdater::build_branch`.

When the prefix is unchanged, we just need a single bitwise memcpy. When the prefix shrinks or grows, we need to do bitwise memcpy in a loop, but it should still be fast. Copying an uncompressed chunk would also be a single bitwise memcpy.

I think this should simplify and speed up the code substantially.

PR #586, opened 2024-11-27
fix: bitwise_memcpy keeps untouched bits

gabriele-0201:

PR #585, opened 2024-11-27
opt: add branch builder push_chunk

gabriele-0201:

PR #569, opened 2024-11-19
opt: BranchNode::set_prefix uses bitwise_memcpy

gabriele-0201:

PR #587, opened 2024-11-27
opt: add binary search in branch updater

gabriele-0201:
This is the first working version of the introduction of binary search within the branch_stage. I'm sure there is still a long way to go for further optimization. At the same time, I have been working on this for weeks and would be happy to receive feedback and use this as a starting point for all future optimizations.

### DIFF
--- a/nomt/src/beatree/ops/update/branch_updater.rs
+++ b/nomt/src/beatree/ops/update/branch_updater.rs
@@ -1,10 +1,12 @@
-use std::cmp::Ordering;
-use std::sync::Arc;
+use std::{ops::Range, sync::Arc};
 
 use crate::beatree::{
     allocator::PageNumber,
     branch::{self as branch_node, BranchNode, BranchNodeBuilder, BRANCH_NODE_BODY_SIZE},
-    ops::bit_ops::{prefix_len, separator_len},
+    ops::{
+        bit_ops::{prefix_len, separator_len},
+        find_key_pos,
+    },
     Key,
 };
 use crate::io::PagePool;
@@ -14,16 +16,39 @@ use super::{
 };
 
 pub struct BaseBranch {
-    pub node: Arc<BranchNode>,
-    pub iter_pos: usize,
+    node: Arc<BranchNode>,
+    low: usize,
 }
 
 impl BaseBranch {
-    fn next_key(&self) -> Option<Key> {
-        if self.iter_pos >= self.node.n() as usize {
-            None
+    pub fn new(node: Arc<BranchNode>) -> Self {
+        BaseBranch { node, low: 0 }
+    }
+
+    // Try to find the given key starting from `self.low` up to the end.
+    // Returns None if `self.low` is already at the end of the node,
+    // or if there are no keys left bigger than the specified one.
+    // If there are available keys in the node, then it returns the index
+    // of the specified key with the boolean set to true or the index containing
+    // the first key bigger than the one specified and the boolean set to false.
+    fn find_key(&mut self, key: &Key) -> Option<(bool, usize)> {
+        if self.low == self.node.n() as usize {
+            return None;
+        }
+
+        let (found, pos) = find_key_pos(&self.node, key, Some(self.low));
+
+        if found {
+            // the key was present return its index and point to the right after key
+            self.low = pos + 1;
+            return Some((true, pos));
+        } else if pos == self.low {
+            // there are no keys left bigger than the specified one
+            return None;
         } else {
-            Some(self.key(self.iter_pos))
+            // key was not present, return and point to the smallest bigger key
+            self.low = pos;
+            return Some((false, pos));
         }
     }
 
@@ -34,15 +59,20 @@ impl BaseBranch {
     fn key_value(&self, i: usize) -> (Key, PageNumber) {
         (self.key(i), self.node.node_pointer(i).into())
     }
-
-    fn advance_iter(&mut self) {
-        self.iter_pos += 1;
-    }
 }
 
+// BranchOp used to create a new node starting off a possible base node.
+//
+// `Update` and `KeepChunk` refers only to compressed separator within the base node.
 enum BranchOp {
+    // Separator and its page number
     Insert(Key, PageNumber),
-    Keep(usize, usize),
+    // Contains the position at which the separator is saved in the base,
+    // along with the updated page number
+    Update(usize, PageNumber),
+    // Contains a range of separators that will be transfered unchanged
+    // in the new node from the base
+    KeepChunk(usize, usize),
 }
 
 pub enum DigestResult {
@@ -84,10 +114,24 @@ impl BranchUpdater {
 
     /// Ingest a key and page number into the branch updater.
     pub fn ingest(&mut self, key: Key, pn: Option<PageNumber>) {
-        self.keep_up_to(Some(&key));
-        if let Some(pn) = pn {
-            self.ops.push(BranchOp::Insert(key, pn));
-            self.bulk_split_step(self.ops.len() - 1);
+        // keep all elements that are skipped looking for `key`
+        let res = self.keep_up_to(Some(&key));
+
+        match (res, pn) {
+            // UNWRAP: if the item has been found it must be a base node
+            (Some(pos), Some(pn))
+                if pos < self.base.as_ref().unwrap().node.prefix_compressed() as usize =>
+            {
+                // a compressed separator has been changed
+                self.ops.push(BranchOp::Update(pos, pn));
+                self.bulk_split_step(self.ops.len() - 1);
+            }
+            (_, Some(pn)) => {
+                // a new key or a previous uncompressed separator has been updated
+                self.ops.push(BranchOp::Insert(key, pn));
+                self.bulk_split_step(self.ops.len() - 1);
+            }
+            _ => (),
         }
     }
 
@@ -111,7 +155,7 @@ impl BranchUpdater {
             self.split(new_branches)
         } else if self.gauge.body_size() >= BRANCH_MERGE_THRESHOLD || self.cutoff.is_none() {
             let node = self.build_branch(&self.ops[last_ops_start..], &self.gauge);
-            let separator = self.op_key(&self.ops[last_ops_start]);
+            let separator = self.op_first_key(&self.ops[last_ops_start]);
             new_branches.handle_new_branch(separator, node, self.cutoff);
 
             self.ops.clear();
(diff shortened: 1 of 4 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/587?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#587** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/587?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/587?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#586** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/586?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#569** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/569?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#585** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/585?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

nomt/src/beatree/ops/update/branch_updater.rs
```diff
@@ -34,15 +59,22 @@ impl BaseBranch {
...
-    fn advance_iter(&mut self) {
-        self.iter_pos += 1;
-    }
 }
 
+// BranchOp used to create a new node starting off a possible base node.
+//
+// `Update` and `KeepChunk` refers to compressed separator present into
+// the base node. Every uncompressed separator present in the base or
+// that will be inserted in the new node will be represented as an `Insert` operation.
```
rphmeier, line 68: this sentence doesn't quite parse!

nomt/src/beatree/ops/update/branch_updater.rs
```diff
@@ -337,63 +493,125 @@ impl BranchUpdater {
...
-                (BranchOp::Keep(pos, _), None) if *pos < base_prefix_compressed => {
-                    pending_chunk = Some((*pos, pos + 1))
+        // This second phase of joining Update and KeepChunk into a unique update chunk is performed
+        // for two reasons:
+        //
+        // 1. It could often happen that the sequence of KeepChunk are interleaved by Update with only a change
+        // in the node pointers
+        // 2. To avoid keeping all the update information within the BranchOp::KeepChunk because it would require
+        // further allocations
+        let apply_chunk =
```
rphmeier, line 503: I don't understand the underlying motivation for this piece of code. It looks like it adds a lot of complexity.

Was performance too slow with it? Is it causing a bug? It looks to me like we could just remove this entire piece of logic and save about 100LoC with dense conditionals. Can you elaborate on reason (2) stated here - all the original data for `KeepChunk` is stored within the base branch, so what causes additional allocations? The signature of push_chunk (base, start, end, ops) should work fine with that.

Our happy-path is

nomt/src/beatree/ops/update/branch_updater.rs
```diff
@@ -337,63 +493,125 @@ impl BranchUpdater {
...
+                    BranchOp::Insert(_, _) => {
+                        apply_chunk(
+                            &mut builder,
+                            pending_keep_chunk.take().unwrap(),
+                            pending_ops_range.take().unwrap(),
+                        );
+                    }
+                    BranchOp::KeepChunk(from, to) => {
+                        // UNWRAP: pending_keep_chunk has just been checked to be Some
+                        let range = pending_keep_chunk.as_mut().unwrap();
```
rphmeier, line 554: Is this branch even possible? Under what circumstances would `keep_up_to` give us two `KeepChunk`s in a row?

rphmeier: Ah...deletions!

nomt/src/beatree/ops/update/branch_updater.rs
```diff
@@ -34,15 +59,22 @@ impl BaseBranch {
...
+// `Update` and `KeepChunk` refers to compressed separator present into
+// the base node. Every uncompressed separator present in the base or
+// that will be inserted in the new node will be represented as an `Insert` operation.
 enum BranchOp {
+    // Separator and its page number
     Insert(Key, PageNumber),
-    Keep(usize, usize),
+    // Contains the position at which the separator is saved in the base,
+    // along with the updated page number
+    Update(usize, PageNumber),
```
rphmeier, line 74: LeafUpdater also has binary search but doesn't have the `Update`. 

I definitely see the motivation for this (swapping page numbers should be cheap, if the key doesn't change, and the separators can be completely copied with a normal `memcpy`).

nomt/src/beatree/ops/update/branch_updater.rs
```diff
@@ -507,6 +808,86 @@ impl BranchGauge {
...
+
+            t = self.total_separator_lengths(p) + base_compressed_separators_len
+
+            // both prefix and total separatos bit len is inherited from the base
+            //p = base.node.prefix_len() as usize;
+            //t = base.node.view().raw_separators_data(from, to).bit_len;
+        }
+
+        let r = branch_node::body_size(p, t, self.n + n_items);
+        //assert_eq!(exptected, r);
```
rphmeier, line 887: should this be removed?

rphmeier (commented):
This looks generally pretty good.

I'm wary of the creeping complexity emerging in the branch updater, which doesn't pair well with a lack of additional test coverage.

For example, we have now in `build_branch` pretty big conditional towers handling a lot of edge cases, and I doubt these are covered in full by our existing unit tests (integration tests notwithstanding).

I'm fine with merging this stack to preserve some velocity, but please follow up with a heavy pass for refactoring and additional testing. We need to devote attention to keep this code manageable and correct.

So while this code does solve the immediate issue of vastly reducing the number of comparisons, and so should be faster, I will note that conditionals metaphorically pour sugar in the gas tank of a modern pipelined CPU - I expect that the branch predictor will get fairly confused and this might add quite a lot of overhead.

nomt/src/beatree/ops/update/branch_updater.rs
```diff
@@ -239,52 +375,72 @@ impl BranchUpdater {
...
+                    let n_items = split_keep_chunk(
+                        self.base.as_ref().unwrap(),
+                        &left_gauge,
+                        &mut self.ops,
+                        split_point,
+                        midpoint,
+                        BRANCH_NODE_BODY_SIZE,
+                    );
+
+                    if n_items == 0 {
```
rphmeier, line 415: I'm not clear why this is necessary; isn't `split_keep_chunk` meant to encapsulate this?

nomt/src/beatree/ops/update/branch_updater.rs
```diff
@@ -137,66 +183,147 @@ impl BranchUpdater {
...
-                    .push(BranchOp::Keep(base_node.iter_pos, separator_len));
+            let base_compressed_end = std::cmp::min(to, base.node.prefix_compressed() as usize);
 
-                base_node.advance_iter();
-                self.bulk_split_step(self.ops.len() - 1);
-            } else {
-                base_node.advance_iter();
-                break;
-            }
+            (from, base_compressed_end, to, found)
```
rphmeier, line 224: this permits `from == base_compressed_end`, leading to us pushing an empty chunk.

nomt/src/beatree/ops/update/branch_updater.rs
```diff
@@ -137,66 +183,147 @@ impl BranchUpdater {
...
+                            if left_n_items > 0 {
+                                // KeepChunk has been successfully split, the left chunk is now able to fit
+                                recursive_on_split = true;
+                                true
+                            } else {
+                                // Not even a single element was able to fit.
+                                if self.gauge.body_size() < BRANCH_MERGE_THRESHOLD {
+                                    // We can stop prefix compression and separate the first
+                                    // element of the keep_chunk into its own.
+                                    self.ops[op_index] = BranchOp::KeepChunk(from + 1, to);
```
rphmeier, line 310: what about 0/1-sized chunks? Are they permitted? what are the invariants around `KeepChunk`?

nomt/src/beatree/ops/update/branch_updater.rs
gabriele-0201: > LeafUpdater also has binary search but doesn't have the Update

That's totally right, and I think it could technically be added, but in that case, I don't think it would be worth it. The optimization takes place from the fact that the same `bitwise_memcpy` can be called on a bigger amount of bytes, assuming that the added computation is faster than calling `bitwise_memcpy` multiple times. I have the gut feeling that something like this could give a positive trade-off only for updated values that keep the same size. But this would require checks to make sure that the size is the same. Second, if the size changes, then a similar amount of byte shift would be required compared to what the current `Insert` and `KeepChunk` approach is doing.

Different from the branches, where the "values" are always 4-byte page numbers

nomt/src/beatree/ops/update/branch_updater.rs
gabriele-0201: No, this could become clearer if the function were renamed `try_split_keep_chunk`, as it accepts both a target and a limit that cannot be exceeded. The function's sole purpose is to split a chunk, which could potentially fail. For example, if the first item in the chunk requires prefix compression of the new node to stop.

This logic is also present in `bulk_split_step`: attempt to split a chunk; if it fails, change the first element of `KeepChunk` to an `Insert` and repeat the loop, potentially leading to the scenario where `left_gauge.stop_prefix_compression();` is reached.

nomt/src/beatree/ops/update/branch_updater.rs
gabriele-0201: Absolutely yes, thanks for funding out the bug! You have already addressed the problem in #597!

nomt/src/beatree/ops/update/branch_updater.rs
gabriele-0201: 0-sized chunks should never be created, but the line of code you just commented on, as well as the previous comment, are two examples of possible cases of 0-sized chunk creations. I will correct this behavior in a follow-up, along with an explanation of the invariant `KeepChunk`

rphmeier:
### Merge activity

* **Dec 10, 6:13 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/587).

PR #596, opened 2024-12-05
fix: update clear bit in page_diff

gabriele-0201:
Sometimes, a page was properly created by the PageWalker, but then in subsequent commits an empty page was found.

For this to occur, two conditions needed to be satisfied:
1. A leaf must have been placed at the last level of a page, resulting in children in the next page.
   For example, a value with a key of 0b00000001_00000000_... which currently resides at the 6th level of the first page.
2. A new leaf with a common prefix of at least 15 bits needed to be introduced.
   For instance, 0b00000001_00000001_...

During the update, a new subtree would be created below the existing leaf, involving the following steps:
1. Clearing the children of the existing leaf (setting the 'cleared' bit to 1 in the relative page_diff)
2. Establishing a new page to store both leaves and their children
3. Writing all internal nodes up to the node where the existing leaf was stored (on the same page where the previous children were present)

The issue arises as step one clears the PageDiff where the children were stored, and step three writes new internal nodes to the same page and page diff without removing the cleared bit. As a result, bitbox reads the cleared bit in the page_diff and ends up deleting the page instead of saving the updated version with the necessary internal nodes.

In summary, if a page initially containing only two children at the first layer were transformed into a page with solely internal nodes, bitbox would not save it.

The solution is straightforward: ensure to use clear bits when eliminating everything, but if writing occurs again, remove the clear bit.

However, there is another tricky part: the changes on the page themselves still need to happen even if a page is being cleared because later on, the same page could be reused. Leave garbage in the nodes would lead to errors. That's why I didn't just put a `set_cleared(false)` within `PageDiff::set_changed(...)`.

Tests will be a part of a follow-up to being able to effectively conduct other tests and benchmarks in the meanwhile!

Closes THR-50

### DIFF
--- a/nomt/src/merkle/page_walker.rs
+++ b/nomt/src/merkle/page_walker.rs
@@ -603,17 +603,22 @@ impl<H: NodeHasher> PageWalker<H> {
         node: Node,
     ) {
         let node_index = self.position.node_index();
-        if self.position.is_first_layer_in_page() {
-            let cleared =
-                node == TERMINATOR && self.sibling_node(write_pass.downgrade()) == TERMINATOR;
-            self.stack.last_mut().unwrap().diff.set_cleared(cleared);
-        }
+        let sibling_node = self.sibling_node(write_pass.downgrade());
 
+        // UNWRAP: if a node is being set, then a page in the stack must be present
         let stack_top = self.stack.last_mut().unwrap();
         stack_top
             .page
             .set_node(&self.page_pool, write_pass, node_index, node);
-        stack_top.diff.set_changed(node_index);
+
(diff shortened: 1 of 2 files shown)


gabriele-0201:
* **#596** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/596?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/596?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

rphmeier (approved):

PR #603, opened 2024-12-06
refactor(beatree): introduce a read transaction type

rphmeier:
This introduces a basic read transaction type (not currently used) that is a first step towards enabling a safe asynchronous API for beatree. The immediate motivations are:
  1. The removal of leaf children requires us to query many leaves asynchronously during merkle update.
  2. Likewise, the rollback log requires us to do the same in order to fetch previous values.

A less immediate but also useful goal is read transactions being used in `Session`s to enable parallel VMs to operate with fewer threads. This is groundwork for that future.

Read transactions don't block commit or `finish_sync` (leaving room for "async sync" in the future), but do block `sync` from beginning. We can expect that sync will _start_ at a natural moment where there are no outstanding read transactions (after a session is committed). When a sync will end is less predictable, and leaving some space for pipelining will be beneficial.

### DIFF
--- a/nomt/src/beatree/mod.rs
+++ b/nomt/src/beatree/mod.rs
@@ -2,10 +2,12 @@ use allocator::{PageNumber, Store, StoreReader, FREELIST_EMPTY};
 use anyhow::{Context, Result};
 use branch::BRANCH_NODE_SIZE;
 use crossbeam_channel::Receiver;
+use imbl::OrdMap;
+
 use leaf::node::MAX_LEAF_VALUE_SIZE;
 use nomt_core::trie::ValueHash;
-use parking_lot::{ArcMutexGuard, Mutex, RwLock};
-use std::{collections::BTreeMap, fs::File, mem, path::Path, sync::Arc};
+use parking_lot::{ArcMutexGuard, Condvar, Mutex, MutexGuard, RwLock};
+use std::{fs::File, mem, path::Path, sync::Arc};
 use threadpool::ThreadPool;
 
 use crate::io::{fsyncer::Fsyncer, IoHandle, IoPool, PagePool};
@@ -27,6 +29,7 @@ pub type Key = [u8; 32];
 
 #[derive(Clone)]
 pub struct Tree {
+    read_transaction_latch: ReadTransactionLatch,
     shared: Arc<RwLock<Shared>>,
     sync: Arc<Mutex<Sync>>,
 }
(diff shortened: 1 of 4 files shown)


rphmeier:
* **#603** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/603?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/603?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

nomt/src/beatree/mod.rs
```diff
@@ -196,7 +223,15 @@ impl Tree {
         let bbn_store;
         let io_handle;
         {
+            // Wait for all outstanding read transactions to conclude. This ensures they won't
+            // be invalidated by any destructive changes.
+            //
+            // Once this guard drops, it will be safe to initiate a new read transaction.
+            let read_tx_guard = read_transaction_latch.block_until_unique();
```
pepyakin, line 230: I spent some time and I think I don't understand why this only blocks here but imediatelly released after we got the write. Why block then at all?

nomt/src/beatree/mod.rs
```diff
@@ -185,8 +207,13 @@ impl Tree {
     /// The returned received indicates that all eviction has completed and it is safe to swap the
     /// index. This receiver must be blocked on before finishing sync.
     ///
+    /// Blocks until all outstanding read transactions have concluded.
```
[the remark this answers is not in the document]
rphmeier: Which two? Do you mean this and the line below?

nomt/src/beatree/mod.rs
rphmeier: It ensures that no new read transactions will be created until the write lock on `self.shared` is released in any case.

It would be fine to hold `read_tx_guard` until the end of this scope as well, of course. But not blocking would lead to a race, where we `block_until_unique` but then a new read transaction is created by another thread before we can acquire `shared.write()`.

nomt/src/beatree/mod.rs
```diff
@@ -401,3 +437,71 @@ impl SyncController {
...
+        ReadTransactionLatch {
+            inner: Arc::new(ReadTransactionLatchInner {
+                read_transactions: Mutex::new(0),
+                cvar: Condvar::new(),
+            }),
+        }
+    }
+
+    fn release_one(&self) {
+        *self.inner.read_transactions.lock() -= 1;
```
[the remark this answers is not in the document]
rphmeier: I can add a check, but there is nothing to do but panic.

nomt/src/beatree/mod.rs
```diff
@@ -401,3 +437,69 @@ impl SyncController {
...
+        // the `add_one` call when the read transaction was created.
+        self.inner.read_transactions.lock().checked_sub(1).unwrap();
+        self.inner.cvar.notify_one();
+    }
+
+    fn add_one(&self) {
+        *self.inner.read_transactions.lock() += 1;
+    }
+
+    fn block_until_unique(&self) -> MutexGuard<usize> {
```
pepyakin, line 495: this blocks until X gets unique.

What is X here?

nomt/src/beatree/mod.rs
pepyakin: Sure, but why is it important to exclude the new readers only for the duration of this scope?

PR #604, opened 2024-12-06
opt: write raw operates on page

pepyakin:
This is one of the first steps on the way to reduce the proliferation of FatPage across the codebase. 

The plan here is to gradually switch to Page in the IO subsystem. In the near future, Write and WriteRaw will be merged together.

### DIFF
--- a/nomt/src/io/page_pool.rs
+++ b/nomt/src/io/page_pool.rs
@@ -76,6 +76,14 @@ impl FatPage {
     pub fn as_mut_ptr(&self) -> *mut u8 {
         self.page.as_mut_ptr()
     }
+
+    /// Returns underlying [`Page`].
+    ///
+    /// Care must be taken that the returned page is not used after `self` is dropped. However,
+    /// failing to do so will not cause safety bugs.
+    pub fn page(&self) -> Page {
+        self.page.clone()
+    }
 }
 
 impl Deref for FatPage {
--- a/nomt/src/beatree/branch/node.rs
+++ b/nomt/src/beatree/branch/node.rs
@@ -8,7 +8,7 @@ use crate::{
         ops::{bit_ops::bitwise_memcpy, get_key},
         Key,
     },
-    io::{FatPage, PagePool},
+    io::{page_pool::Page, FatPage, PagePool},
 };
 
 // Here is the layout of a branch node:
@@ -52,6 +52,10 @@ impl BranchNode {
 
     pub fn as_slice(&self) -> &[u8] {
         &*self.page
+    }
+
+    pub fn page(&self) -> Page {
+        self.page.page()
     }
 
     pub fn view(&self) -> BranchNodeView {
--- a/nomt/src/beatree/ops/update/branch_stage.rs
+++ b/nomt/src/beatree/ops/update/branch_stage.rs
@@ -8,8 +8,7 @@ use crate::beatree::{
     index::Index,
     Key,
 };
-
-use crate::io::{IoCommand, IoHandle, IoKind, PagePool, PAGE_SIZE};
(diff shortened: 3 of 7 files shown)


pepyakin:
* **#606** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/606?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#605** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/605?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#604** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/604?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/604?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

nomt/src/io/page_pool.rs
```diff
@@ -76,6 +76,14 @@ impl FatPage {
     pub fn as_mut_ptr(&self) -> *mut u8 {
         self.page.as_mut_ptr()
     }
+
+    /// Returns underlying [`Page`].
+    ///
+    /// Care must be taken that the returned page is not used after `self` is dropped. However,
+    /// failing to do so will not cause safety bugs.
```
rphmeier, line 83: not to say this function should be `unsafe` (it shouldn't; dereferencing the pointer is the unsafe bit) but this can cause data races

deuszx: I'm sure there's an explanation for doing it like this but if you warn about usage after self is dropped - why not return reference and disallow cloning/copying entirely? I understand you want to act on the underlying (in `Page`) data in a way that requires `self` (in `IoKind::WriteRaw`) but maybe there's an alternative? Like using a reference there as well - it seems that only `as_ptr()` is used there which on itself needs `&self`

pepyakin: > I'm sure there's an explanation for doing it like this but if you warn about usage after self is dropped - why not return reference and disallow cloning/copying entirely? I understand you want to act on the underlying (in Page) data in a way that requires self (in IoKind::WriteRaw) but maybe there's an alternative? Like using a reference there as well - it seems that only as_ptr() is used there which on itself needs &self

I am not really sure what you mean here, @deuszx.

WriteRaw cannot really use references, because those require specifying lifetimes, and the best you can do there is a 'static lifetime. So you need a kind of a smart-pointer. FatPage is a kind of one.

We want to get rid of it, because it incurs a non-trivial performance overhead, specifically because it is based off an Arc (to the page pool).

So the overall idea is to get rid of the FatPage altogether. There are several techniques and observations that can help with that. 

1. One of them is that often write requests come in batches and perhaps it'd be benefitial to release them all at once carrying around a single Arc to page pool.
2. The pages always return to the io sender and thus they can be released using a page pool which likely hangs around. We just need to make sure that page pool lives enough to not outlive any of the pages... doesn't seem too complicated.

rphmeier: >  is this a suggestion or just a heads up

Sorry, I wasn't clear there at all. The comment reads as "However, failing to [not use the page after `self is dropped] will not cause safety bugs". But using the page after `self` is dropped definitely can cause safety bugs, since it is a use-after-free. So editing that or removing it if it's not the intended meaning would be more accurate.

rphmeier (approved):

rphmeier:
### Merge activity

* **Dec 10, 6:21 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/604).

PR #610, opened 2024-12-11
feat(beatree): add a database iterator implementation

rphmeier:
This PR introduces an iterator over the database. Our initial use-case is looking up leaf node preimages instead of the leaf children we currently store in bitbox. 

Though on first inspection it might not seem like our intended use-case requires an iterator, it actually does in the presence of overlays - the degenerate case we might need to handle where there _were_ an arbitrary amount N of keys with a prefix but N-1 of them were deleted in the overlay. This requires us to iterate all leaves where the item might be, at least until we find it.

The goal of this iterator type is to be pretty lightweight. Creating an iterator does an arc clone of the index and both memory overlays, and does a single index/branch lookup to find the first leaf.

The iterator does not handle leaf fetching itself, but instead exposes a `needed_leaves` function which itself returns an iterator that can be used to determine all the upcoming leaves. This `needed_leaves` has a 'static lifetime, so it can be pocketed away in future use-cases where we aggressively pre-fetch the next needed leaves over large iterations.

Future use-cases might include:
  - Exporting the state in full (useful in rehashing / regenerating the DB)
  - Exposing iteration capabilities in the `Session`.

All exposed types have the 'static lifetime.

### DIFF
--- a/nomt/src/beatree/iterator.rs
+++ b/nomt/src/beatree/iterator.rs
@@ -1,0 +1,543 @@
+//! Database iterators over the Beatree.
+
+// TODO: remove this once used.
+#![allow(unused)]
+
+use std::{
+    cmp::Ordering,
+    ops::{Range, RangeFrom},
+    sync::Arc,
+};
+
+use imbl::ordmap::{Iter as OrdMapIter, OrdMap};
+use nomt_core::trie::ValueHash;
+
+use super::{
+    allocator::PageNumber,
+    branch::node::{get_key, BranchNode},
+    index::Index,
+    leaf::node::LeafNode,
+    Key, ReadTransaction, ValueChange,
+};
+
+/// An iterator over the state of the beatree at some particular point.
+///
+/// This combines the in-memory overlays with the state of the leaf pages on the disk.
+/// This iterator does not handle the fetching of pages internally, but instead provides the needed
+/// page numbers. The [`ReadTransaction`]` will provide facilities for dispatching I/O with a
+/// provided handle.
+///
+/// This is not a normal Rust iterator, due to its need to block. Furthermore, it is a streaming
+/// iterator which does not clone or copy its outputs, rather returning them as borrowed. This means
+/// that the standard iterator combinators can't be used with it, making it less versatile than a
+/// typical Rust iterator.
+pub struct BeatreeIterator {
+    memory_values: StagingIterator,
+    leaf_values: LeafIterator,
+}
+
+impl BeatreeIterator {
+    pub(super) fn new(tx: &ReadTransaction, start: Key, end: Option<Key>) -> Self {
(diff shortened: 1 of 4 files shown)


rphmeier:
* **#610** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/610?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/610?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

nomt/src/beatree/iterator.rs
```diff
@@ -0,0 +1,543 @@
...
+        BeatreeIterator {
+            memory_values: StagingIterator::new(tx, start, end),
+            leaf_values: LeafIterator::new(tx.bbn_index.clone(), start, end),
+        }
+    }
+
+    /// Get the next value from the iterator.
+    ///
+    /// This may either return a value or a marker indicating that it is blocked on a page load.
+    /// The page to load can be determined with [`BeatreeIterator::next_pns`].
```
gabriele-0201, line 50: Probably a non-updated comment, should be something like 'the next pages to be loaded are provided by `needed_page`'

nomt/src/beatree/iterator.rs
```diff
@@ -0,0 +1,543 @@
...
+        // because we first have to advance the state of the iterator and then return a borrow
+        // from the new state.
+        //
+        // we need to first point the state at the next item and then borrow it out of the advanced
+        // state.
+        //
+        // result: None => None, Some(false) => blocked, Some(true) => take last.
+        let (new_state, result) =
+            match std::mem::replace(&mut self.state, LeafIteratorState::Done { last: None }) {
+                s @ LeafIteratorState::Done { .. } => (s, None),
```
gabriele-0201, line 244: nice usage of this syntax!

gabriele-0201 (approved):
I found it a little complex to review this pr because every time I had a doubt about something, or if I didn't understand it, I had to check the follow-up pr containing the bug fixes and minor refactorings. It could have been possible to fold the two prs together, or maybe just split the tests into a separate follow-up pr. Aside from that, and modulo fixes, the PR is great!

rphmeier:
> I found it a little complex to review this pr because every time I had a doubt about something, or if I didn't understand it, I had to check the follow-up pr containing the bug fixes and minor refactorings

Yes, sorry - I hadn't considered that you might be looking at bugs already fixed.

rphmeier:
### Merge activity

* **Dec 16, 12:48 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/610).

PR #618, opened 2024-12-13
refactor: move I/O handle ownership out of bitbox pageloader

rphmeier:
Moving control of the I/O handle out of the `PageLoader` enables it to be used for other things, like querying the beatree for leaf preimages.


rphmeier:
* **#618** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/618?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/618?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

gabriele-0201 (approved):
The PR perfectly achieves the goal of abstracting out the ownership of the `IoHandle`, but I found the way it does it a little bit confusing.

Now the page load logic is spread between the `PageLoad` struct (who's responsible to abstract the probe logic and send the correct read io command) and the seeker who needs to handle completion on its own. Two possible simple alternatives are to change the `PageLoader` to be more of a `PageProber` and let the seeker to send and receive commands, or just move back the completion logic into the `PageLoader` with the trade-off of some more LoC.

Is there a specific reason why the completion logic was moved out of the `PageLoader`?

rphmeier (commented):
> Is there a specific reason why the completion logic was moved out of the PageLoader?

I don't think the premise here is correct.

The completion logic was moved from `PageLoadCompletion` into `PageLoad`. I removed `PageLoadCompletion` because it didn't make sense here any more, since it could only be produced by having the `PageLoader` wrap the I/O handle. Then the home for `try_complete` could have either gone into `PageLoader` or `PageLoad`. I opted for `PageLoad`, but either one would be fine.

The reason I didn't want the `PageLoader` to wrap the I/O handle was so the `Seeker` could use the same I/O handle for bitbox and beatree.

I think what you suggest would look almost exactly the same, maybe we can chalk it up to style.

rphmeier:
### Merge activity

* **Dec 16, 12:48 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/618).

PR #624, opened 2024-12-17
refactor: add BranchOpsTracker

gabriele-0201:
The following includes a significant refactor of the branch updater. The main goal was to encapsulate all the logic for branch operation handling elsewhere, in this case in the `BranchOpsTracker`.

I'm not attached to specific names, so suggestions are welcome. I find it extremely interesting how the `BranchUpdater` has been greatly simplified, and as a result, the module's source of errors seems to be drastically reduced.

It is a substantial refactor, and I wasn't able to split it across multiple smaller pull requests. Looking at how many lines changed it seems completely new code, but it isn't. Just most of the logic that was spread around the `BranchUpdter` is logically moved into the `BranchOpsTracker`. The only real change is in how bulk split is handled. It is not something that is checked on each ingestion, but it is only attempted during the `digest` method.

### DIFF
--- a/nomt/src/beatree/ops/update/branch_ops.rs
+++ b/nomt/src/beatree/ops/update/branch_ops.rs
@@ -1,0 +1,969 @@
+use std::ops::Deref;
+
+use crate::beatree::{
+    branch::node::{self, get_key, BRANCH_NODE_BODY_SIZE},
+    ops::{
+        bit_ops::separator_len,
+        update::{
+            branch_updater::{BaseBranch, BranchGauge},
+            BRANCH_MERGE_THRESHOLD,
+        },
+    },
+    Key, PageNumber,
+};
+
+// BranchOp used to create a new node starting off a possible base node.
+//
+// `Update` and `KeepChunk` refers only to compressed separator within the base node.
+pub enum BranchOp {
+    // Separator and its page number
+    Insert(Key, PageNumber),
+    // Contains the position at which the separator is saved in the base,
+    // along with the updated page number
+    Update(usize, PageNumber),
+    // Contains a range of separators that will be transfered unchanged
+    // in the new node from the base.
+    //
+    // These are always items which were prefix-compressed.
+    KeepChunk(KeepChunk),
+}
+
+// KeepChunk represents a sequence of separators contained in a branch node.
+// `start` and `end` represents the separators range where `end` is non inclusive.
+// 0-Sized chunks are not allowed, thus `start > end` must always hold.
+// `sum_separator_lengths` is the sum of the separator lengths of each separator
+// represented by the chunk itself.
+#[derive(Debug, Clone, Copy)]
+pub struct KeepChunk {
+    pub start: usize,
+    pub end: usize,
+    pub sum_separator_lengths: usize,
+}
+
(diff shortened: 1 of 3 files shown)


nomt/src/beatree/ops/update/branch_ops.rs
```diff
@@ -0,0 +1,965 @@
...
+        // After ingesting all those ops, the bulk split is expected to be initiated.
+        assert_eq!(n_splits, expected_bulk_split);
+        // The gauge needs to contain all the ops not present in the previous split.
+        assert_eq!(ops_tracker.gauge.n(), n_keys - n_items_in_splits);
+        // Operations are expected to be drainded.
+        assert_eq!(ops_tracker.ops.len(), n_keys - n_items_in_splits);
+    }
+
+    #[test]
+    fn exrtact_ops_until_only_inserts() {
```
rphmeier, line 444: typo here and elsewhere `exrtact`

nomt/src/beatree/ops/update/branch_ops.rs
```diff
@@ -0,0 +1,965 @@
...
+            };
+
+            self.ops.insert(index, BranchOp::KeepChunk(left_chunk));
+            self.ops[index + 1] = BranchOp::KeepChunk(right_chunk);
+        }
+        left_chunk_n_items
+    }
+}
+
+pub struct BranchOps<'a> {
```
rphmeier, line 370: docs on public items please!

nomt/src/beatree/ops/update/branch_ops.rs
```diff
@@ -0,0 +1,965 @@
...
+                gauge,
+            ))
+        } else {
+            self.valid_gauge = true;
+            self.gauge = gauge;
+            None
+        }
+    }
+
+    pub fn prepare_merge_ops(&mut self, base: Option<&BaseBranch>) {
```
rphmeier, line 293: docs

nomt/src/beatree/ops/update/branch_ops.rs
```diff
@@ -0,0 +1,965 @@
...
+
+        // Replace with Insert if:
+        // 1. Prefix compression is stopped.
+        // 2. Update op is referring to an uncompressed separator.
+        if base.node.prefix_compressed() as usize <= pos || self.gauge.prefix_compressed.is_some() {
+            self.replace_with_insert(Some(base), self.ops.len() - 1);
+        }
+    }
+
+    pub fn push_chunk(&mut self, base: &BaseBranch, start: usize, end: usize) {
```
rphmeier, line 93: docs

rphmeier (approved):
This refactor looks pretty nice. The way that bulk split is handled now also looks very good, as a simple loop. Fewer gauge checks in the critical path of ingestion seems like it'd be pretty efficient.

nomt/src/beatree/ops/update/branch_ops.rs
gabriele-0201: Weird. I'm using `cargo-spellcheck` to ensure I'm not leaving typos or more obvious errors, but it didn't flag those. I will correct them now obv!

PR #633, opened 2024-12-18
refactor: integrate async overflow lookup into beatree API

rphmeier:


nomt/src/beatree/mod.rs
```diff
@@ -6,6 +6,7 @@ use imbl::OrdMap;
 
 use leaf::node::MAX_LEAF_VALUE_SIZE;
 use nomt_core::trie::ValueHash;
+use ops::overflow::AsyncReader as AsyncOverflowReader;
```
pepyakin, line 9: I find those rebindings really confusing. I think we already discussed this, but here is a rough recap:

You see `AsyncOverflowReader`. You think it's likely a struct. there is no definition of that in vicinity, so you just grep for `struct AsyncOv...`. No results. Then you look for `AsyncOverflowReader` and after multiple ^Gs you end up on this little line that renames it for you.

IMO much better to disambiguate using `overflow::AsyncReader`: disambiguates perfectly, doesn't raise questions.

nomt/src/beatree/mod.rs
```diff
@@ -626,19 +637,82 @@ impl AsyncLeafLoad {
...
+                }
+
+                res
+            }
+        }
     }
 }
 
+/// Metadata related to an overflow page fetch.
+pub struct OverflowMetadata(usize);
```
pepyakin, line 714: Not really clear what kind of metadata is that. Not even sure if that's a metadata, if you think about it.

pepyakin (approved):
LGTM nits only

pepyakin:
### Merge activity

* **Dec 19, 8:09 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/633).

nomt/src/beatree/mod.rs
rphmeier: well, what do you recommend calling it?

pepyakin: I don't really have a good name, not now. Something that mentions that it represents an inflight fetch and that it is an index etc

PR #648, opened 2024-12-20
feat: overlay type definitions and operations

rphmeier:
This defines the `Overlay` types. I tried to make it passably fast at an algorithmic level and will follow up with integrations!

### DIFF
--- a/nomt/src/overlay.rs
+++ b/nomt/src/overlay.rs
@@ -1,0 +1,265 @@
+//! A utility for managing in-memory overlays.
+//!
+//! This module exposes two types. The user-facing [`Overlay`] is opaque and frozen. The internal
+//! [`LiveOverlay`] is meant to be used within a session only.
+//!
+//! Overlays contain weak references to all their ancestors. This allows ancestors to be dropped or
+//! committed during the lifetime of the overlay. Importantly, this means that memory is cleaned
+//! up gracefully as overlays are dropped and committed.
+//!
+//! However, creating a new [`LiveOverlay`] requires the user to provide strong references to each
+//! of the ancestors which are still alive. It is still the user's responsibility to ensure
+//! that all live ancestors are provided, or else data will go silently missing.
+//!
+//! Looking up a value in an overlay does not have a varying cost. First there is a look-up in an
+//! index to see which ancestor has the data, and then a query on the ancestor's storage is done.
+//!
+//! Creating a new overlay is an O(n) operation in the amount of changes relative to the parent,
+//! both in terms of new changes and outdated ancestors.
+
+#![allow(dead_code)]
+
+use crate::{beatree::ValueChange, page_cache::Page, page_diff::PageDiff};
+use nomt_core::{page_id::PageId, trie::KeyPath};
+
+use std::collections::HashMap;
+use std::sync::{Arc, Weak};
+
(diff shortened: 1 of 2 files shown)


rphmeier:
* **#648** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/648?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

nomt/src/overlay.rs
```diff
@@ -0,0 +1,264 @@
...
+//! This module exposes two types. The user-facing [`Overlay`] is opaque and frozen. The internal
+//! [`LiveOverlay`] is meant to be used within a session only.
+//!
+//! Overlays contain weak references to all their ancestors. This allows ancestors to be dropped or
+//! committed during the lifetime of the overlay. Importantly, this means that memory is cleaned
+//! up gracefully as overlays are dropped and committed.
+//!
+//! However, creating a new [`LiveOverlay`] requires the user to provide strong references to each
+//! of the ancestors which are still alive. It is still the user's responsibility to ensure
+//! that all live ancestors are provided, or else data will go silently missing.
```
gabriele-0201, line 12: I struggle to see how a user can create a chain of `Overlay` with some missing `Overlay`, leading to silently miss out on some data. I have a some ideas based on how you envisioned the usage in the follow-up pr.

The user will be able to start a session and update NOMT with an `Overlay` as an output (first session created, so no overlays are available to be submitted). From now on, new overlays can be created again from the same previous state or on top of fresh new overlays. Each new overlay will contain a vector of weak references to all ancestors, making two things possible:

1. Upon the creation of `LiveOverlay`, the provided Overlay iterator is checked against the vector of weak references contained in the parent of the new overlay.
2. During the commit of an overlay, some logic could be added to ensure that an overlay is not committed twice or that all ancestors have already been committed. Something like storing a vector of all weak references to childs of an overlay could make this possible (modulo some efficiency, I didn't think about yet to this aspect)

That is to say, the comment seems not strictly correct to me, the user will probably be able to not commit an overlay and thus may silently lose some data (but probably this could be solved in some way). However, they will not be able to silently create an overlay on top of a broken chain of overlays, because an error will be thrown as a result.

gabriele-0201 (approved):
Modulo some doubts on comments this seems to be a great starting point to manage `Overlay`s. I like the `imbl` and `Map + Vector` approach to handle a quick indexing of the overlays!

nomt/src/overlay.rs
rphmeier: In this iteration of the code, it's possible to just omit some of the ancestors.

For example, if you have a chain of overlays `[A B C D]`, and none of them have been committed, it would be allowed by the API for the user to create a new overlay with only ancestors `[C D]`. None of the value/page changes from `[A B]` would be reflected in the `LiveOverlay`, leading to inconsistencies. I address this now in the latest revision of the following PR, where all uncommitted ancestors must be provided.

rphmeier:
### Merge activity

* **Jan 14, 2:29 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/648).

PR #649, opened 2024-12-20
feat: update* functions in NOMT API

rphmeier:
These are all unimplemented at the moment, but gives a hint of what the API would look like.

I do think there is a combinatorial explosion (especially if we split out `sync` in the future). This is just an initial approach to get started and then we can do a full API pass later to make it nice.

### DIFF
--- a/nomt/src/lib.rs
+++ b/nomt/src/lib.rs
@@ -17,6 +17,7 @@ use nomt_core::{
     trie::{InternalData, NodeHasher, NodeHasherExt, ValueHash, TERMINATOR},
     trie_pos::TriePosition,
 };
+use overlay::{LiveOverlay, OverlayMarker};
 use page_cache::PageCache;
 use parking_lot::Mutex;
 use store::Store;
@@ -26,7 +27,9 @@ use store::Store;
 pub use nomt_core::proof;
 pub use nomt_core::trie::{KeyPath, LeafData, Node, NodePreimage};
 pub use options::{Options, PanicOnSyncMode};
+pub use overlay::{InvalidAncestors, Overlay};
 pub use store::HashTableUtilization;
+
 // beatree module needs to be exposed to be benchmarked
 #[cfg(feature = "benchmarks")]
 #[allow(missing_docs)]
@@ -58,6 +61,8 @@ pub type Value = Vec<u8>;
 struct Shared {
     /// The current root of the trie.
     root: Node,
+    /// The marker of the last committed overlay. `None` if the last commit was not an overlay.
+    last_commit_marker: Option<OverlayMarker>,
 }
 
 /// A witness that can be used to prove the correctness of state trie retrievals and updates.
@@ -232,7 +237,10 @@ impl<T: HashAlgorithm> Nomt<T> {
             page_cache,
             page_pool,
             store,
-            shared: Arc::new(Mutex::new(Shared { root })),
+            shared: Arc::new(Mutex::new(Shared {
+                root,
+                last_commit_marker: None,
+            })),
             session_cnt: Arc::new(AtomicUsize::new(0)),
             metrics,
             _marker: std::marker::PhantomData,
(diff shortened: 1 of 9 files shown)


rphmeier:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/649?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#649** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/649?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/649?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#648** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/648?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

nomt/src/lib.rs
```diff
@@ -291,14 +310,49 @@ impl<T: HashAlgorithm> Nomt<T> {
...
+        unimplemented!()
     }
 
-    /// Commit the transaction and returns the new root.
+    /// Commit the changes from this overlay to the underlying database.
+    ///
+    /// This assumes no sessions are active and will panic otherwise.
+    ///
+    /// The caller should take care to commit overlays in order and not to commit conflicting
+    /// overlays. Otherwise, the database will be left in an inconsistent state.
```
rphmeier, line 346: I don't love this, and I'm considering enforcing overlays to be committed in sequence. It seems super error prone to just corrupt the DB silently.

nomt/src/lib.rs
```diff
@@ -259,17 +261,34 @@ impl<T: HashAlgorithm> Nomt<T> {
...
+            .unwrap()
+    }
+
+    /// Creates a new [`Session`] object on top of a series of in-memory overlays.
+    ///
+    /// This will fail if the first overlay was not created on top of the previous overlays.
+    /// Providing an empty iterator will give the same result as `begin_session`.
+    ///
+    /// It is the user's responsibility to ensure that all uncommitted overlays are provided.
+    /// Failing to provide all uncommitted overlays will lead to incomplete proofs or corruption.
```
rphmeier, line 277: Likewise here. This also seems super error-prone.

I don't know if there's a legitimate reason to ever omit these, and it would be better to fail with a nice error instead.

gabriele-0201: > Failing to provide all uncommitted overlays

Do you mean that calling this function with an empty iterator of overlays while they are instead non-committed will lead to an error? Otherwise, given how `Overlay`s are created, I don't see how it would be possible to provide an incorrect iterator of `Overlay`s that would not be detected by `LiveOverlay::new(overlays)`, leading to an error being returned and not to incorrect overlays being committed

nomt/src/lib.rs
gabriele-0201: I agree that enforcing ordering, maybe with something like a new in-memory atomic `commit_seqn`, would reduce the potential for errors

gabriele-0201 (approved):
It's okay how the API now handles both situations, with overlays being on top or not! I still need to address the doubt about building new sessions on top of a chain of overlays

nomt/src/lib.rs
rphmeier: It's the same as discussed in the last issue: the user could omit some uncommitted overlays. Previously, we just checked that they are ancestors, not that the list of ancestors is complete. However, that's fixed now and no longer relevant.

rphmeier:
### Merge activity

* **Jan 14, 2:29 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/649).

PR #651, opened 2024-12-23
refactor: add deep Clone to FatPage

rphmeier:

PR #652, opened 2024-12-23
refactor: merklize into copied pages

rphmeier:
There are two reasons to do this:
  1. Perf. We copy the pages anyway during sync and it's currently a single-threaded bottleneck. Doing it during `commit` where we have more threads would be best.
  2. Overlays. Overlays can't mutate pages in place, so they need to be copied.

I suspect that this will actually be _faster_ than `master` for commit-concurrency > 1. However, I will also benchmark for commit-concurrency = 1 (note: I remove the copies during writeout only further upstack, so the benchmarks will be there)

This seems to obviate the read/write pass stuff, so if perf holds up we should likely just remove those in follow-up PRs.

### DIFF
--- a/nomt/src/merkle/page_walker.rs
+++ b/nomt/src/merkle/page_walker.rs
@@ -53,9 +53,8 @@ use nomt_core::{
 
 use crate::{
     io::PagePool,
-    page_cache::{Page, PageCache, PageCacheShard, ShardIndex},
+    page_cache::{Page, PageCache, PageCacheShard, PageMut},
     page_diff::PageDiff,
-    rw_pass_cell::{ReadPass, RegionContains, WritePass},
 };
 
 /// The output of the page walker.
@@ -63,18 +62,22 @@ pub enum Output {
     /// A new root node.
     ///
     /// This is always the output when no parent page is supplied to the walker.
-    Root(Node, Vec<(PageId, PageDiff)>),
+    Root(Node, Vec<UpdatedPage>),
     /// Nodes to set in the bottom layer of the parent page, indexed by the position of the node
     /// to set.
     ///
     /// This is always the output when a parent page is supplied to the walker.
-    ChildPageRoots(Vec<(TriePosition, Node)>, Vec<(PageId, PageDiff)>),
+    ChildPageRoots(Vec<(TriePosition, Node)>, Vec<UpdatedPage>),
 }
 
-struct StackItem {
-    page_id: PageId,
-    page: Page,
-    diff: PageDiff,
+/// A page which was updated during the course of modifying the trie.
+pub struct UpdatedPage {
+    /// The ID of the page.
+    pub page_id: PageId,
+    /// An owned copy of the page, including the modified nodes.
+    pub page: PageMut,
+    /// A compact diff indicating all modified slots in the page.
+    pub diff: PageDiff,
 }
 
 /// The source of pages for the page walker.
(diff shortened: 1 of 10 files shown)


rphmeier:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/652?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#652** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/652?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#651** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/651?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

nomt/src/merkle/page_walker.rs
```diff
@@ -816,42 +789,50 @@ mod tests {
...
         walker.advance_and_replace(
-            &mut write_pass,
             TriePosition::new(),
             vec![
                 (key_path![0, 1, 0, 1, 1, 0], val(1)),
                 (key_path![0, 1, 0, 1, 1, 1], val(2)),
             ],
         );
 
+        match walker.conclude() {
```
rphmeier, line 801: this test was actually silently broken in master for a couple reasons.

1. it broke the contract of the page walker, because it called `advance_and_replace` with a position that was a sub-position of a previous `advance_and_replace` call.
2. it tested that there were 3 modified pages, however, in the previous version of the test there were only two pages that were being modified (root + 1 child page). the third page came from a broken invariant causing `build_stack` to reload a page which was already pushed to the `updated_pages` vector.

the changes in this PR exposed the problems with this test so I fixed them here.

nomt/src/merkle/page_walker.rs
```diff
@@ -359,29 +351,32 @@ impl<H: NodeHasher> PageWalker<H> {
...
+                        page: PageMut::pristine_empty(),
+                        diff: PageDiff::default(),
+                    }
+                } else if self
+                    .updated_pages
+                    .last()
+                    .map_or(false, |d| d.page_id == child_page_id)
+                {
+                    // UNWRAP: just checked
+                    self.updated_pages.pop().unwrap()
```
rphmeier, line 366: I am fairly certain this is never reachable after the removal of leaf children. The reason this case might've existed before was:
 1. we begin to replace a leaf at the bottom of a page with 2 leaves at the layer below. 
 2. its leaf children are on the beginning of the next page, so we remove them (page gets pushed onto `updated_pages`)
 3. we then traverse `down` into that page and need to get it from `updated_pages`.
 
However, now we don't have this case. We never remove something from a page and then re-enter that page.

gabriele-0201: I also agree with that and with the reason why it was previously required. This could be proven by the fact that `down()` is only called in `replace_terminal(..)`. If it satisfies the fact that all new elements of a subtree are provided in one batch, then there is no way to do what you just explained

gabriele-0201 (approved):
Not needing the read-write pass anymore increases a lot code readability, maintaining clear separations between which worker will update which pages!

nomt/src/merkle/page_walker.rs
pepyakin: Should we put an `unreachable!` guard here then?

pepyakin (approved):

pepyakin:
### Merge activity

* **Jan 2, 7:23 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/652).

PR #658, opened 2025-01-02
expose sync_seqn

pepyakin:
Makes the sync_seqn number accessible. Helps to detect whether a commit succeeded or not.

Closes THR-86

### DIFF
--- a/nomt/src/lib.rs
+++ b/nomt/src/lib.rs
@@ -258,6 +258,12 @@ impl<T: HashAlgorithm> Nomt<T> {
         self.store.load_value(path)
     }
 
+    /// Returns the current sync sequence number.
+    #[doc(hidden)]
+    pub fn sync_seqn(&self) -> u32 {
+        self.store.sync_seqn()
+    }
+
     /// Creates a new [`Session`] object, that serves a purpose of capturing the reads and writes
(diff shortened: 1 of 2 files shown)


gabriele-0201 (approved):

nomt/src/lib.rs
```diff
@@ -258,6 +258,12 @@ impl<T: HashAlgorithm> Nomt<T> {
         self.store.load_value(path)
     }
 
+    /// Returns the current sync sequence number.
+    #[doc(hidden)]
```
gabriele-0201, line 262: Why is the documentation hidden? Is the goal to make it visible only to people who want to make more complex integrations and thus read nomt's source code?
Because if the goal is to have it pub only for use by `superagent`, but ideally not by external users, it may be better to put it behind a feature flag

pepyakin: > Why is the documentation hidden? 

This is not what it says. Instead, it means that this is an exposed but not documented method, or equally, not a part of public API. That means that people can still use it and I don't have a problem with that. Those same people should be aware though that using non-documented APIs can lead to breakage.

The feature flag is another way to solve this problem I agree. But I don't think that all this complexity that comes with it is justified.

PR #632, opened 2024-12-18
feat(beatree): add an asynchronous overflow value fetcher

rphmeier:
This isn't used (in this PR) but can be used in:
  1. (immediately) rollback to avoid blocking on overflow values
  2. (later) any asynchronous session stuff we add later.

PR #659, opened 2025-01-06
feat: tighten read of overflow node

gabriele-0201:
An overflow cell can be part of the Cell array within each leaf, and it's formed in the following way:
  + overflow Cell: (u64, u256, [NodePointer]) | semantically, (value_size, value_hash, [NodePointer])

The array of NodePointer can be at maximum long `MAX_OVERFLOW_CELL_NODE_POINTERS = 15` elements.

The format fo an overflow page is:
+ n_pointers: u16
+ n_bytes: u16
+ pointers: [PageNumber; n_pointers]
+ bytes: [u8; n_bytes]

There are multiple reads and allocations that depend on the data written on disk. Firstly, the `value_size` present in the overflow_cell can be extremely big, as mentioned  [[https://github.com/thrumdev/nomt/pull/632#discussion_r1891948861][here]]. A vector is preallocated based on the `value_size` to contain the value, but if it is too big the pre-allocation could lead to resource exhaustion. To prevent this, `MAX_OVERFLOW_VALUE_SIZE` is used in both encoding and decoding of the `overflow_cell`.

`MAX_OVERFLOW_VALUE_SIZE` is currently set to 4TiB.

Different for both `n_pointers` and `n_bytes` which are used directly to access the `FatPage` and thus a panic will happen if the reads happen to be out of bounds.

Closes THR-62


gabriele-0201:
* **#659** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/659?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/659?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

nomt/src/beatree/ops/overflow.rs
```diff
@@ -93,18 +93,25 @@ pub fn decode_cell<'a>(raw: &'a [u8]) -> (usize, [u8; 32], impl Iterator<Item =
     assert!(raw.len() >= 8 + 4 + 32);
     assert_eq!(raw.len() % 4, 0);
 
-    let value_size = u64::from_le_bytes(raw[0..8].try_into().unwrap());
+    let value_size = u64::from_le_bytes(raw[0..8].try_into().unwrap()) as usize;
+    // SAFETY: values bigger than MAX_OVERFLOW_VALUE_SIZE are not allowed.
+    assert!(value_size < MAX_OVERFLOW_VALUE_SIZE);
```
pepyakin, line 98: To align with the comment and the logic in `encode_cell` this should be:

```suggestion
    assert!(value_size <= MAX_OVERFLOW_VALUE_SIZE);
```

pepyakin (approved):
Lgtm!

I would vote to reduce the MAX_OVERFLOW_VALUE_SIZE to something reasonable.

gabriele-0201:
Fine for me reducing `MAX_OVERFLOW_VALUE_SIZE`! What value seems more reasonable for you? 1TiB? or even less?

pepyakin:
Here you are essentially setting the maximum value size limit. I don't think we ever need 1 TiB. I think we don't even need 1 GiB. 512 MiB should be enough for everyone.

gabriele-0201:
> 512 MiB should be enough for everyone

I also agree that it should be enough, but at the same time, I didn't want to prevent some esoteric usage of nomt from happening. However, I'm absolutely not pushing for keeping 4TiB. I would be okay with 1GiB or 512MiB and then, if needed, raise it!

nomt/src/beatree/ops/overflow.rs
```diff
@@ -93,18 +93,25 @@ pub fn decode_cell<'a>(raw: &'a [u8]) -> (usize, [u8; 32], impl Iterator<Item =
     assert!(raw.len() >= 8 + 4 + 32);
     assert_eq!(raw.len() % 4, 0);
 
-    let value_size = u64::from_le_bytes(raw[0..8].try_into().unwrap());
+    let value_size = u64::from_le_bytes(raw[0..8].try_into().unwrap()) as usize;
+    // SAFETY: values bigger than MAX_OVERFLOW_VALUE_SIZE are not allowed.
```
pepyakin, line 97: There is no unsafe in sight. PANIC?

nomt/src/beatree/ops/overflow.rs
```diff
@@ -413,7 +422,7 @@ mod tests {
 
     impl Arbitrary for ValidOverflowCell {
         fn arbitrary(g: &mut Gen) -> Self {
-            let value_size = usize::arbitrary(g) & 0x7FFFFFFF; // Keep sizes reasonable
+            let value_size = usize::arbitrary(g) & 0x1FFFFFFF; // Keep sizes below 512MiB
```
pepyakin, line 425: Wouldn't it be better to derive this constant from the size itself?

nomt/src/beatree/ops/overflow.rs
gabriele-0201: Thanks for letting me notice this! The word 'safety' should not be used here at all

nomt/src/beatree/ops/overflow.rs
gabriele-0201: You are totally right, and making it dynamic relative to the constant is really easy. I was rushing to publish it. Sorry!

PR #660, opened 2025-01-06
feat: tighten read of branch node

gabriele-0201:
The structure of a branch node is the following:
+ bbn_pn: u32
+ n: u16
+ prefix_compressed: u16
+ prefix_len: u16
+ cells: u16[n]
+ prefix: bitvec[prefix_len]
+ separators: bitvec
+ padding
+ node_pointers: leaf_node_page_number[n]

The unsafety in the module is driven by reading `cells` and `node_pointers` as slices with the number of elements given by `n`. Ensuring that `n` leads to a read within the bounds of the `FatPage` will make the read panic instead of some other errors.

Every other read passes through an access of the `FatPage`, so any out of bounds access will lead to a panic.

Closes THR-63

### DIFF
--- a/nomt/src/beatree/branch/node.rs
+++ b/nomt/src/beatree/branch/node.rs
@@ -122,6 +122,11 @@ impl BranchNode {
     }
 
     fn cells_mut(&mut self) -> &mut [[u8; 2]] {
+        let cells_end = BRANCH_NODE_HEADER_SIZE + self.n() as usize * 2;
+        assert!(cells_end < BRANCH_NODE_BODY_SIZE);
+
+        // SAFETY: This creates a slice of length 2 * N starting at index BRANCH_NODE_HEADER_SIZE.
+        // This is ensured to be within the bounds of the page by the assertion above.
         unsafe {
             std::slice::from_raw_parts_mut(
                 self.page[BRANCH_NODE_HEADER_SIZE..BRANCH_NODE_HEADER_SIZE + 2].as_ptr()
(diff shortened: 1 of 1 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/660?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#660** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/660?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/660?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#659** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/659?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

rphmeier (commented):
conditional panics in the fast-path should be free, given the branch predictor and 2-way pipeline execution, but just in case we might want to revisit this by validating at branch node creation time instead (one-time cost rather than each-time cost)

rphmeier (approved):

rphmeier:
### Merge activity

* **Jan 8, 5:42 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/660).

gabriele-0201:
> but just in case we might want to revisit this by validating at branch node creation time instead

This would not cover the case of loading a wrong bbn file where a bit has flipped, modifying `n`.

Because if we just rely on 'we read what we wrote,' then the `assert` is pointless because intrinsically, the creation logic will create a correct node. By trusting the encoding, we could assume that `n` is always correct. But the main purpose of this assert, in my opinion, is to protect against reading something we are not 100% sure we wrote, for example, the bbn file being modified between two runs.

PR #681, opened 2025-01-10
test: fix fill_and_empty tests

gabriele-0201:
make fill_and_empty reproducible and effectively
write and read all generated keys

PR #682, opened 2025-01-10
fix: first leaf mut have a separator of all 0

gabriele-0201:
What this pr implements is a function to enforce that the separator of the first leaf in the beatree is all 0.

The reason for opting for a new function is to avoid adding complexity to the leaf_stage logic. Carrying around information about the first leaf would have required a lot of modifications here and there. Now, instead, that rule is enforced by only the new function and the `unwrap_or` within the `LeafUpdater::separate` function. Additionally, this edge case is very unlikely, and the function will, for the most part, return after the first check. The goal was to minimize the impact on the critical path and reduce code complexity while maintaining the rule's validity.

Closes THR-91

### DIFF
--- a/nomt/src/beatree/ops/update/leaf_stage.rs
+++ b/nomt/src/beatree/ops/update/leaf_stage.rs
@@ -175,6 +175,7 @@ pub fn run(
     }
 
     output.leaf_changeset.sort_by_key(|(k, _)| *k);
+    enforce_first_leaf_separator(&mut output.leaf_changeset, bbn_index);
     Ok(output)
 }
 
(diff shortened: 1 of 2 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/682?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#683** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/683?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#682** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/682?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/682?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#681** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/681?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

nomt/src/beatree/ops/update/leaf_stage.rs
```diff
@@ -208,6 +209,96 @@ fn apply_worker_changes(
...
+
+    // Initially set `maybe_new_first` to be the second leaf in the previous state.
+    // It is an option because there may have been only one leaf in the previous state.
+    // `maybe_new_first` always refers to leaves that were present previously.
+    let mut separator = [0; 32];
+    let mut maybe_new_first: Option<([u8; 32], PageNumber)>;
+
+    // Iterate over `leaf_changeset` until a non deleted item with a separator
+    // smaller than `maybe_new_first` is found. Meanwhile, if we find `maybe_new_first`
+    // to be deleted, update it with the following one in the previous state.
```
rphmeier, line 245: could this loop be simplified?

It seems that you can find the next leaf with a simple approach:
  1. Find the 2nd leaf (first branch, 2nd separator) in the index.
  1. If it exists, then compare it with the first leaf in the changeset. If the one in the changeset is earlier, then use that. 
  1. If it doesn't exist, then take the first leaf in the changeset. If the changeset is otherwise empty, then the DB is empty.

gabriele-0201: > If it exists, then compare it with the first non-deleted leaf in the changeset

within step 2, here it seems that you are losing a couple of possibilities. The second previous leaf, the one extracted in what you described as step 1, could be either eliminated or modified within the current changeset.

What I'm doing is what you described with only the addition that step two is a loop, because the second leaf could be eliminated so we need to check the third and so on.

Could this logic be implemented more nicely? probably! I will give it a second look and I will try to simplify it!

rphmeier: Makes sense!

rphmeier:
### Merge activity

* **Jan 14, 2:26 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/682).

PR #698, opened 2025-01-17
feat(torture): crash after average commit time

gabriele-0201:
The following feature was very easy to implement and highly effective as it significantly increased crash occurrences during sync. 
  
With `trickfs`, this issue will be secondary, as it will allow actions such as crashing when writing to a specific memory area or file. Therefore, I'm fine if you prefer to close this PR and wait for `trickfs`!"

PR #697, opened 2025-01-17
feat(torture): add commit crash bias

gabriele-0201:

PR #696, opened 2025-01-17
feat(torture): add `WorkloadParams` to cli

gabriele-0201:

PR #700, opened 2025-01-17
feat(torture): sample state after commit

gabriele-0201:


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/700?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#702** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/702?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#701** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/701?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#700** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/700?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/700?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#698** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/698?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#697** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/697?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#696** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/696?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

torture/src/supervisor/workload.rs
```diff
@@ -162,7 +162,10 @@ impl WorkloadState {
         // - Different power of two sizes.
         // - Change it to be a non-even.
         let len = if self.rng.gen_bool(self.biases.overflow) {
-            32 * 1024
+            // TODO: 1332 is MAX_LEAF_VALUE_SIZE, so 2000 is enough.
+            // It has been decreased because `32 * 1024` causes the channel to
+            // crash when the agent was sending back the value.
```
pepyakin, line 167: - Are those comments meant to be addressed as TODO within this PR or in general?
- If in general, then it is going to be confusing when viewed not in the context of this diff/commit.
- Can you share more details on the nature of this crash? By the channel you mean the comms socket, correct? What do you mean by "channel crash"? Do you mean a single value of `32 * 1024` bytes won't fit the channel or do you mean multiple values pushed into the channel make it overflow?

torture/src/supervisor/workload.rs
```diff
@@ -353,15 +384,39 @@ impl Workload {
...
+    }
 
-        // TODO: Exercise should sample the values from the state.
+    async fn ensure_changeset_reverted(
+        &self,
+        rr: &comms::RequestResponse,
+        changeset: &Vec<KeyValueChange>,
+    ) -> anyhow::Result<()> {
+        // Given a reverted changeset, we need to ensure that each modified/deleted key
+        // is equal to its previous state and that each new key is not avaible.
```
pepyakin, line 405:
```suggestion
        // is equal to its previous state and that each new key is not available.
```

pepyakin (approved):

torture/src/supervisor/workload.rs
gabriele-0201: Not in this PR but in a follow-up.

What should I have done differently? I noticed a bug relate to this value, it was causing torture to fail. I added a TODO in the comments to address it later on. The goal of the comment was to provide context for the change. To be fair, I expected to fix this immediately after this pr, but other tasks took priority on my todo list, and I am only starting to work on it now.

> Can you share more details

unfortunately not yet, I'm still trying to figure out the nature of the error

pepyakin: Ok, gotcha. I am curious about the pt.3 because I am worried that this is just a symptom and we are just slapping a patch on it here.

> The goal of the comment was to provide context for the change. 

Yes, that's fine. I am just nitting the way it's phrased and the information it conveys. The particular instance does not matter, I don't think it can introduce any damage and I will certainly sleep soundly with this thing in in the codebase. I am just trying to set a bar for communcation here.

Now with that out of the way, consider that somebody unfamiliar with this change reads this code (like an external contributor or yourself in some months). The part that mentions `32 * 1024`, IMO, is confusing because it refers to something that is not in the current version of the codebase but in the previous. People are mostly dealing with files at the recent versions of the repo and all the other context is basically lost unless you explicitly link it.

Also, IMO, it lacks some detail. We decreased the value to 2000 but then the natural question would be "does it test the multiple overflow page scenario?". The answer, afaiu, is no. This might be an important peace of information to address!

So, my suggestion for this particular case would be to:

1. mention that this is a workaround for a bug.
1. describe the behavior you work around or maybe just mention that it's unclear what is happening but with workaround it works.
1. remove the `32 * 1024` part
1. mention that we should sacrifice testing the multi-page overflows because of this and that (if my assumption turns out to be correct of course).

gabriele-0201: > I am worried that this is just a symptom and we are just slapping a patch on it here.

definitely it is just a patch here. It seems to be completely separated by this feature

> external contributor or yourself in some months ... is confusing because it refers to something that is not in the current version of the codebase

You are 100% right about this , it confusing right now. It would have been a lot less problematic if it was followed immediately with the fix, as I intended initially. 

Thanks for noticing this! And all the suggestions! I will follow them and change the code accordingly!

PR #717, opened 2025-01-21
refactor(torture): probabilities handling

gabriele-0201:
Define a `rollback-bias` argument which defines the probability
of executing a rollback instead of a commit.
Add then `rollback-crash-bias` and `commit-crash-bias`,  two
arguments to define the probability of causing to crash respectively
the rollback or the commit when they would be performed.

This is based on what @pepyakin pointed out [here](https://github.com/thrumdev/nomt/pull/706#discussion_r1922495335)

### DIFF
--- a/torture/src/supervisor/cli.rs
+++ b/torture/src/supervisor/cli.rs
@@ -63,29 +63,29 @@ pub struct WorkloadParams {
     #[arg(long = "random-size")]
     pub random_size: bool,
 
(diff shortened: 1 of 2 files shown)


pepyakin:
### Merge activity

* **Jan 28, 5:24 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/717).

PR #730, opened 2025-01-23
fix: label pages in WAL recovery

rphmeier:
WAL recovery wrote the nodes into pages, but not the page ID.

That meant that fresh pages which are recovered only from the WAL would not be probeable and would be treated as misprobes, silently losing data. The fix is to set the page ID on WAL recovery.

### DIFF
--- a/nomt/src/bitbox/mod.rs
+++ b/nomt/src/bitbox/mod.rs
@@ -442,6 +442,9 @@ fn recover(
                 }
                 page_diff.unpack_changed_nodes(&changed_nodes, &mut page);
 
+                // Label the page.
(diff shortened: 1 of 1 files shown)


rphmeier:
* **#730** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/730?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/730?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #738, opened 2025-01-24
fix(seglog): choose segment correctly when pruning front

rphmeier:
The logic for finding the segment to prune was broken.

If you have 2 segments covering ranges [1, 5] and [6, 10] and prune for segment 7, you want to take the second one, not the first. But the `max <= end_live` check would give you the first segment (and then fail because it doesn't actually contain segment 7).

Most of the time, we only have 1 segment so this wasn't triggering.

### DIFF
--- a/nomt/src/seglog/mod.rs
+++ b/nomt/src/seglog/mod.rs
@@ -354,7 +354,7 @@ impl SegmentedLog {
         let mut seg_index = self.segments.len() - 1;
         while seg_index > 0 {
             let segment = &self.segments[seg_index];
-            if segment.max <= new_end_live {
+            if segment.min <= new_end_live {
                 break;
             }
             seg_index -= 1;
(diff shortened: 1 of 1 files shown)


rphmeier:
* **#738** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/738?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

nomt/src/seglog/mod.rs
```diff
@@ -1323,6 +1323,34 @@ mod tests {
...
+
+        h.new_segment(3)?
+            .write_record(7, &[4u8; 10])?
+            .write_record(8, &[5u8; 10])?
+            .write_record(9, &[6u8; 10])?
+            .write()?;
+
+        let (mut log, _) = h.open_log(default_max_segment_size(), 1, 9)?;
+        assert_eq!(log.live_range(), (RecordId::from(1), RecordId::from(9)));
+        log.prune_front(5.into())?;
```
rphmeier, line 1349: with the old logic, this was choosing segment 1 (3 <= 5) and not segment 2 (6 > 5). Then failing because it couldn't find record 5 in segment 1.

gabriele-0201 (approved):

PR #781, opened 2025-02-05
refactor: put node hasher in charge of labeling nodes and determining kinds

rphmeier:
1. Export a couple more things from the NOMT crate root that make it possible to implement a NodeHasher
2. Make determining the node kind a function of the NodeHasher, and give the NodeHasher the responsibility of labeling nodes.

### DIFF
--- a/examples/witness_verification/src/main.rs
+++ b/examples/witness_verification/src/main.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use nomt_core::{
     proof,
-    trie::{LeafData, NodeHasher},
+    trie::{self, LeafData, Node, NodeHasher, NodeKind, TERMINATOR},
 };
 
 // Hash nodes using blake3. The Hasher below will be utilized in the witness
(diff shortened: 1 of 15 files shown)


rphmeier:
* **#781** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/781?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/781?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

examples/witness_verification/src/main.rs
```diff
@@ -10,8 +10,36 @@ use nomt_core::{
...
+    }
+
+    fn node_kind(node: &Node) -> NodeKind {
+        if node[0] >> 7 == 1 {
+            NodeKind::Leaf
+        } else if node == &TERMINATOR {
+            NodeKind::Terminator
+        } else {
+            NodeKind::Internal
+        }
```
pepyakin, line 42: As this is an example it will be good to mention why we are doing all of that stuff.

1. Why are implementing a node hasher here ourselves? Just to demonstrate. Although arguably we might need to use a separate example for that.
2. Why is the code doing what it does? We should probably mention what's the deal with that but direct the user for more details to the rustdoc?

rphmeier: I will follow up with a bit more refactoring. But actually we do it here because `Blake3Hasher` isn't exposed from `nomt-core`.

PR #790, opened 2025-02-07
feat(torture): unify exec commit and commit crash

gabriele-0201:
The unification of the two has the side effect of handling
ENOSPC errors also for commits that are expected to crash.

### DIFF
--- a/torture/src/supervisor/workload.rs
+++ b/torture/src/supervisor/workload.rs
@@ -414,101 +414,96 @@ impl Workload {
             }
         }
 
-        if self.state.rng.gen_bool(self.state.biases.commit_crash) {
-            self.exercise_commit_crashing(&rr).await?;
-        } else {
-            self.exercise_commit(&rr).await?;
-        }
+        let should_crash = self.state.rng.gen_bool(self.state.biases.commit_crash);
+        self.exercise_commit(&rr, should_crash).await?;
 
         Ok(())
     }
 
     /// Commit a changeset.
-    async fn exercise_commit(&mut self, rr: &comms::RequestResponse) -> anyhow::Result<()> {
-        trace!("exercising commit");
+    async fn exercise_commit(
+        &mut self,
+        rr: &comms::RequestResponse,
+        should_crash: bool,
+    ) -> anyhow::Result<()> {
+        let should_crash = if should_crash {
+            trace!("exercising commit crash");
+            Some(self.get_crash_delay())
+        } else {
+            trace!("exercising commit");
+            None
+        };
 
+        // Generate a changeset and the associated snapshot
         let (snapshot, changeset) = self.state.gen_commit();
-        let ToSupervisor::CommitResponse { elapsed, outcome } = rr
+        let commit_response = rr
             .send_request(crate::message::ToAgent::Commit(
                 crate::message::CommitPayload {
                     changeset: changeset.clone(),
-                    should_crash: None,
+                    should_crash,
                 },
             ))
-            .await?
-        else {
-            return Err(anyhow::anyhow!("Commit did not execute successfully"));
+            .await?;
+
+        let is_applied = if should_crash.is_some() {
+            let ToSupervisor::Ack = commit_response else {
+                return Err(anyhow::anyhow!("Commit crash did not execute successfully"));
+            };
+
+            self.wait_for_crash().await?;
+
+            // During a commit crash, every type of error could happen.
+            // However the agent will be respawned, so it will just
+            // make sure the changeset was correctly applied or reverted.
+            self.spawn_new_agent().await?;
+            let agent = self.agent.as_ref().unwrap();
+            let rr = agent.rr().clone();
+
+            // Sample the agent to make sure the changeset was correctly applied or reverted.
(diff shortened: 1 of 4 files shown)


gabriele-0201:
* **#792** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/792?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#791** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/791?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#790** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/790?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/790?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

torture/src/supervisor/workload.rs
```diff
@@ -414,43 +414,107 @@ impl Workload {
...
-        else {
-            return Err(anyhow::anyhow!("Commit did not execute successfully"));
+            .await?;
+
+        // Ensure that the outcome is what we expect.
+        if should_crash.is_some() {
+            self.wait_for_crash().await?;
+            let ToSupervisor::Ack = commit_response else {
+                return Err(anyhow::anyhow!("Commit crash did not execute successfully"));
+            };
```
pepyakin, line 453: IMO it's kind of weird that first we wait for the crash and only then check the response 🤷

torture/src/supervisor/workload.rs
```diff
@@ -414,43 +414,107 @@ impl Workload {
...
+
+            self.ensure_outcome_validity(rr, &outcome).await?;
+
+            if matches!(outcome, crate::message::Outcome::Success) {
+                self.n_successfull_commit += 1;
+                self.tot_commit_time += elapsed;
+            }
         };
 
+        // Udpate RequestResponse, the agent could have been respawned.
```
pepyakin, line 472:
```suggestion
        // Update RequestResponse, the agent could have been respawned.
```

although this particular action is often referred as "reload", so a better version would be

```suggestion
        // Reload RequestResponse, the agent could have been respawned.
```

torture/src/supervisor/workload.rs
```diff
@@ -414,43 +414,107 @@ impl Workload {
...
+        let agent = self.agent.as_ref().unwrap();
+        let rr = agent.rr().clone();
+
+        // Sample the agent to make sure the changeset was correctly applied or reverted.
+        let agent_sync_seqn = rr.send_query_sync_seqn().await?;
+        if snapshot.sync_seqn != agent_sync_seqn {
```
pepyakin, line 478: ❗️

Shouldn't we expect, for example, that if the commit succeeded then the sync_seqn should actually be equal to `snapshot.sync_seqn`?

torture/src/supervisor/workload.rs
```diff
@@ -414,43 +414,107 @@ impl Workload {
...
-                    return Err(anyhow::anyhow!("Commit should have failed with ENOSPC"));
+                    return Err(anyhow::anyhow!("Operation should have failed with ENOSPC"));
                 }
             }
-            crate::message::CommitOutcome::StorageFull => {
+            crate::message::Outcome::StorageFull => {
                 // The commit failed due to `ENOSPC`.
                 if !self.enabled_enospc {
-                    return Err(anyhow::anyhow!("Commit should have succeeded"));
+                    return Err(anyhow::anyhow!("Operation should have succeeded"));
```
pepyakin, line 517: If so, then update the comment just above.

pepyakin (approved):

torture/src/supervisor/workload.rs
gabriele-0201: Thanks for noticing this. Definitely, the check for response needs to be placed before waiting for the crash

torture/src/supervisor/workload.rs
gabriele-0201: Yes, we expected that. The current if-else statement doesn't properly cover all expected and unexpected cases. Now that you've let me think more about that, I will update it with:

```rust
        // Sample the agent to make sure the changeset was correctly applied or reverted.
        let agent_sync_seqn = rr.send_query_sync_seqn().await?;
        if snapshot.sync_seqn == agent_sync_seqn {
            // The changeset is expected to be applied, the commit succeeded.
            self.ensure_changeset_applied(&rr, &changeset).await?;
            self.state.commit(snapshot);
        } else if self.state.committed.sync_seqn == agent_sync_seqn {
            // The changeset is expected to be reverted, the commit crashed.
            self.ensure_changeset_reverted(&rr, &changeset).await?;
        } else {
            return Err(anyhow::anyhow!("Unexpected agent sync_seqn"));
        }

```

pepyakin:
### Merge activity

* **Feb 15, 2:32 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/790).

PR #795, opened 2025-02-10
fix(seglog): `pop_front` in memory log when pruning

gabriele-0201:
when `self.shared.max_rollback_log_len` is reached and rollback delta
needs to be pruned from the front, `prune_to_new_start_live` will be the
`RecordId` just after the oldest delta in memory, thus `pop_front` is required.

Closes THR-118

### DIFF
--- a/nomt/src/rollback/mod.rs
+++ b/nomt/src/rollback/mod.rs
@@ -83,6 +83,10 @@ impl InMemory {
         self.log.pop_back()
     }
 
+    fn pop_front(&mut self) -> Option<(RecordId, Delta)> {
+        self.log.pop_front()
+    }
+
     // Returns the total number of deltas, including the staged one.
     fn total_len(&self) -> usize {
         self.log.len()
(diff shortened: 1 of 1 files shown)


gabriele-0201:
* **#795** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/795?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/795?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

pepyakin (approved):
Nice, thanks.

This suggests that we might need to consider to change the reminology from front/back to something like `push_recent` and `pop_oldest`. It's vecdeque who cannot afford this, and we, working in the specific domain, can.

Would you be open to file a test for this case, while your wet caches are warm?

gabriele-0201:
Yes, I will address renaming and tests in a follow up!

PR #797, opened 2025-02-10
feat: wait for shutdown

pepyakin:
This changeset stops `Nomt::drop` from returning until all the IO
workers finished. Also, this only releases flock on the database
directory only after the IO workers finished.

This matters when creating a NOMT instance right after shutting down the
previous one, as might happen upon recovery after poisoning.


gabriele-0201 (commented):
I think that this pull request will be okay after addressing just two things: the possible failure of the assertion I explained in the comment, and also after what we discussed privately, a requalification of the pull request to make its title and description more toward "Wait for all pending I/O requests before dropping"

pepyakin:
### Merge activity

* **Feb 12, 3:29 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/797).

PR #800, opened 2025-02-11
refactor(page_cache): introduce a sub-cache for merkle pages below a fixed depth

rphmeier:
This puts all pages with level <= 3 into a "fixed level cache" hash-map.
Cache eviction will never remove these pages and they can only be removed if the page itself is deleted.

This will be followed up by a PR for prepopulation.

### DIFF
--- a/nomt/src/page_cache.rs
+++ b/nomt/src/page_cache.rs
@@ -14,12 +14,15 @@ use nomt_core::{
     trie::Node,
 };
 use parking_lot::{Mutex, RwLock};
-use std::{fmt, num::NonZeroUsize, sync::Arc};
+use std::{collections::HashMap, fmt, num::NonZeroUsize, sync::Arc};
 
 // Total number of nodes stored in one Page. It depends on the `DEPTH`
 // of the rootless sub-binary tree stored in a page following this formula:
 // (2^(DEPTH + 1)) - 2
 pub const NODES_PER_PAGE: usize = (1 << DEPTH + 1) - 2;
+
+// The fixed number of levels we always preserve in-memory.
+const FIXED_LEVELS: usize = 3;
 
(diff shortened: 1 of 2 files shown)


nomt/src/page_cache.rs
```diff
@@ -14,13 +14,16 @@ use nomt_core::{
...
-use std::{fmt, num::NonZeroUsize, sync::Arc};
+use std::{collections::HashMap, fmt, num::NonZeroUsize, sync::Arc};
 
 // Total number of nodes stored in one Page. It depends on the `DEPTH`
 // of the rootless sub-binary tree stored in a page following this formula:
 // (2^(DEPTH + 1)) - 2
 pub const NODES_PER_PAGE: usize = (1 << DEPTH + 1) - 2;
 
+// The fixed number of levels we always preserve in-memory.
+const FIXED_LEVELS: usize = 3;
```
pepyakin, line 25: This is slightly wrong.

If the root is the 0th level, we still keep 1st, 2nd and the 3rd levels. So the number that figures in the constant is 3 but we keep 4 levels.

nomt/src/page_cache.rs
```diff
@@ -146,11 +149,50 @@ struct CacheShard {
 }
 
 struct CacheShardLocked {
+    // storage for pages in the levels of the tree which we always cache.
+    fixed_level_cache: HashMap<PageId, CacheEntry, FxBuildHasher>,
```
pepyakin, line 153: Nit: "Fixed" is ok, but I wonder if something like pinned would be slightly better?

pepyakin (approved):

pepyakin:
### Merge activity

* **Feb 12, 3:48 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/800).

PR #816, opened 2025-02-18
feat(torture): if agent hangs print its bt

pepyakin:
To aid debugging cases where the agent got deadlocked (or just timed
out), print the backtrace of every thread of the agent process.

This PR removes dumpable(false) so the caveat might apply. We need the agent process to be dumpable because otherwise we cannot trace it and no debugger can be attached.

### DIFF
--- a/torture/src/supervisor/pbt.rs
+++ b/torture/src/supervisor/pbt.rs
@@ -1,0 +1,53 @@
+//! Collection of process backtraces.
+//!
+//! This uses the grug-brain developer approach: just invoke the LLDB or GDB to get the backtrace.
+
+use std::path::Path;
+use tokio::{fs, process::Command};
+use which::which;
+
+pub async fn collect_process_backtrace(filename: &Path, pid: u32) -> anyhow::Result<()> {
+    // Determine which debugger tool to use.
+    let command_str = if which("lldb").is_ok() {
+        lldb(pid)
+    } else if which("gdb").is_ok() {
+        gdb(pid)
+    } else {
+        anyhow::bail!("no lldb or gdb in PATH")
+    };
+
+    // Run the command using a shell
+    let output = Command::new("sh")
+        .arg("-c")
+        .arg(&command_str)
+        .output()
+        .await?;
+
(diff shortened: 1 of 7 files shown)


pepyakin:
* **#816** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/816?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/816?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

torture/src/supervisor/pbt.rs
```diff
@@ -0,0 +1,48 @@
+//! Collection of process backtraces.
+//!
+//! This uses the grug-brain developer approach: just invoke the LLDB or GDB to get the backtrace.
```
rphmeier, line 3: simple but effective :)

torture/src/supervisor/pbt.rs
```diff
@@ -0,0 +1,48 @@
...
+use which::which;
+
+pub async fn collect_process_backtrace(pid: u32) -> anyhow::Result<String> {
+    // Determine which debugger tool to use.
+    let command_str = if which("lldb").is_ok() {
+        lldb(pid)
+    } else if which("gdb").is_ok() {
+        gdb(pid)
+    } else {
+        anyhow::bail!("Neither lldb nor gdb are available in PATH.")
```
rphmeier, line 15: Maybe would be helpful to say _why_ torture is looking for those tools. For example, "attempted to print backtrace of stalled process. ensure lldb or gdb are in PATH"

rphmeier (approved):

torture/src/supervisor/pbt.rs
pepyakin: Actually, we print that up the stack!

pepyakin:
I ran this thing overnight and got a hang. It got printed in the console but it did not fit into the buffer! Changed it to print into a file.

PR #817, opened 2025-02-19
feat(torture): accept custom supervisor workdir

gabriele-0201:


gabriele-0201:
* **#818** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/818?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#817** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/817?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/817?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

torture/src/supervisor/mod.rs
```diff
@@ -156,6 +160,30 @@ pub struct InvestigationFlag {
...
+
+impl WorkloadDir {
+    fn path(&self) -> String {
+        match self {
+            WorkloadDir::TempDir(temp_dir) => temp_dir.path().display().to_string(),
+            WorkloadDir::Dir(p) => p.display().to_string(),
+        }
+    }
+
+    // Will persist the TempDir in memory if called.
```
pepyakin, line 178: At this point why not just make it called as that?
E.g. `ensure_stable_path`.

torture/src/supervisor/workload.rs
```diff
@@ -919,7 +918,7 @@ impl Workload {
     }
 
     /// Return the working directory.
-    pub fn into_workdir(self) -> TempDir {
-        self.workdir
+    pub fn into_workload_dir(self) -> WorkloadDir {
```
pepyakin, line 921: rustdoc

torture/src/supervisor/workload.rs
```diff
@@ -240,7 +240,7 @@ impl WorkloadState {
 /// it behaves.
 pub struct Workload {
     /// Working directory for this particular workload.
-    workdir: TempDir,
+    workload_dir: WorkloadDir,
```
pepyakin, line 243: rustdoc

If you introduce the dichotomy between the working directory and workload directory, then I suggest you do not refer to the latter as "working [workload] directory", since that's a recipe for confusion.

torture/src/supervisor/mod.rs
```diff
@@ -166,20 +194,39 @@ async fn run_workload(
...
+        WorkloadDir::TempDir(tempdir)
+    };
+
+    let mut workload = Workload::new(seed, workload_dir, workload_params, workload_id)?;
     let result = workload.run(cancel_token).await;
     match result {
-        Ok(()) => Ok(None),
+        Ok(()) => {
+            // Clean the persistent db if the workload succeeded.
+            if to_clean_workload_dir {
```
pepyakin, line 220: nit: As it seems to me you already have this distinction indicated by the variant of the WorkloadDir.

Wouldn't it be better to examine the workload dir again and

1. if it is a "TempDir" — then just drop it and its desctructor does its thing
2. if it is a "Dir" — you perform the thing inside this block.

This could be it's own method on `WorkloadDir` eg `WorkloadDir::cleanup(self)` or just `drop` (although in the latter case you will have to swallow).

torture/src/supervisor/mod.rs
```diff
@@ -166,20 +194,39 @@ async fn run_workload(
...
+                let workload_dir = workload.into_workload_dir();
+                std::fs::remove_dir_all(workload_dir.path()).unwrap();
```
pepyakin, line 222: `?`

torture/src/supervisor/mod.rs
```diff
@@ -156,6 +160,30 @@ pub struct InvestigationFlag {
     reason: anyhow::Error,
 }
 
+/// Whether NOMT instances will be built on top of temporary
+/// or user specified directory.
+pub enum WorkloadDir {
+    TempDir(TempDir),
+    Dir(PathBuf),
+}
```
pepyakin, line 168: Nit: This could be more descriptive to reflect what those are exactly, to highlight their distinction. E.g. think about this:

`TempDir` sets up and destroys a temporary directory whereas `Dir` reuses a single one removing files in between the runs.

gabriele-0201: I agree that this documentation is missing. What you said about the `Dir` variant is not true. `WorkloadDir` is different from the working directory specified by the user. The function `run_workload` will use the supervisor's working directory to create the `WorkloadDir::Dir` within it, and a new directory will be created on every run. If the workload succeeds, the directory will be cleaned (this is automatically done by the temp directory), but if it fails, that workload directory will be persistent in memory, and another workload will start in a new directory.

torture/src/supervisor/mod.rs
gabriele-0201: Unfortunately, `TempDir` does not implement `Clone`. To make the code less complex, I decided to keep moving the `TempDir` within the `Workload` struct to avoid dealing with lifetimes or using references to `WorkingDir`, which would have involved even more refactoring. Implementing a simple boolean to track whether the directory would require cleanup or not seemed to be effective and clean.

torture/src/supervisor/workload.rs
gabriele-0201: Totally, this was just an error to not update this doc!

torture/src/supervisor/mod.rs
gabriele-0201: This was a weird usage of `into_workload_dir`, I updated it!

gabriele-0201:
> ... using a single directory obviously precludes that ...

I think this PR may have been misunderstood. The `workdir` (working directory) can be overwritten, and whether or not it is overwritten, it will contain multiple workload directories, allowing for multiple parallel workloads.

Now I pushed multiple changes, waiting for a new review!

torture/src/supervisor/mod.rs
```diff
@@ -157,6 +162,31 @@ pub struct InvestigationFlag {
...
+
+impl WorkloadDir {
+    fn path(&self) -> PathBuf {
+        match self {
+            WorkloadDir::TempDir(temp_dir) => temp_dir.path().into(),
+            WorkloadDir::Dir(p) => p.clone(),
+        }
+    }
+
+    // Will persist the TempDir in memory if called.
```
pepyakin, line 181: 1. rustdoc
2. "in memory" is not really technically accurate, we are talking about [disk] storage here, not about RAM which is typically implied by "memory".

pepyakin (approved):

torture/src/supervisor/mod.rs
gabriele-0201: Totally! I lost this comment along the way! I will update it in a pr!

PR #826, opened 2025-02-24
fix/feat(torture): handle panics

gabriele-0201:
Add utilities to handle panics (`Box<dyn Any + Send>`) and forward
them to the supervisor when they happen within the agent.

### DIFF
--- a/torture/src/agent.rs
+++ b/torture/src/agent.rs
@@ -17,9 +17,12 @@ use tokio_stream::StreamExt as _;
 use tokio_util::codec::{FramedRead, FramedWrite, LengthDelimitedCodec};
 use tracing::trace;
 
-use crate::message::{
-    self, CommitPayload, Envelope, InitOutcome, KeyValueChange, OpenOutcome, OpenPayload, Outcome,
-    RollbackPayload, ToAgent, ToSupervisor, MAX_ENVELOPE_SIZE,
+use crate::{
+    message::{
+        self, CommitPayload, Envelope, InitOutcome, KeyValueChange, OpenOutcome, OpenPayload,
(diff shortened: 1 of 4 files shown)


torture/src/agent.rs
```diff
@@ -303,7 +312,10 @@ impl Agent {
         let nomt = self.nomt.as_ref().unwrap();
 
         // Perform the rollback.
-        let rollback_result = tokio::task::block_in_place(|| nomt.rollback(n_commits));
+        let rollback_result = tokio::task::block_in_place(|| {
+            std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| nomt.rollback(n_commits)))
+                .unwrap_or_else(|e| panic_to_err("Panic in rollback", e))
+        });
```
pepyakin, line 318: Nit: why not extract this into a common function?

torture/src/supervisor/mod.rs
```diff
@@ -48,18 +48,8 @@ pub async fn run() -> Result<()> {
...
-                Some(s.clone())
-            } else {
-                None
-            };
-            if let Some(s) = panic_string {
-                error!("Panic occured: {}", s);
-            } else {
-                error!("Panic occured (no message)");
-            }
+            let panic_string = crate::panic_handling::panic_to_string("Panic occured", panic_box);
```
pepyakin, line 51: The `context` parameter in this call site stands out with it's ambiguity compared to other call sites.

torture/src/agent.rs
gabriele-0201: Fair, it has been extracted now!

torture/src/supervisor/mod.rs
gabriele-0201: That's true, it is more ambiguous than others, To make it clearer, I also update the error message associated with `ExitReason::Error`!

PR #832, opened 2025-02-25
feat(torture): latency injection

gabriele-0201:
The following pr introduces 'Latency Injection' within `Trickfs`. This can be triggered by the `TrickHandle`, and if triggered, every `trickfs` operation will be subject to a scheduled reply after a certain delay.

Notes:
1. In order to deal with latencies, the timeout for waiting for agent replies needed to be increased, as sometimes there were quite big delays.
2. Two of the most crucial `trickfs` operations are read and write, which I tried to keep simple while injecting latencies. For writes, they happen immediately but the reply of things getting written is scheduled. The same applies to reads, where things are copied into the scheduled reply. This latter part introduces some latency by itself, but it does not significantly impact the code with lifetimes or check if memory got modified in the meantime.
3. To keep `torture` reproducible to a certain extent, I added a seed to `trickfs`, which also needs to be specified in all other occasions. Let me know if you have a better approach to handle this while keeping it reproducible.

Closes THR-114

### DIFF
--- a/trickfs/src/latency.rs
+++ b/trickfs/src/latency.rs
@@ -1,0 +1,91 @@
+use std::{
+    collections::VecDeque,
+    sync::mpsc::{self, Receiver, RecvTimeoutError, Sender},
+    time::{Duration, Instant},
+};
+
+use rand::SeedableRng;
+use rand_distr::Distribution;
+
+/// Max possible delay, in micros, used as injected latency.
+const MAX_LATENCY_MICROS: u64 = 1500;
+type Reply = Box<dyn FnOnce() + Send + 'static>;
+
+/// An injector of latencies.
+///
+/// This allows to schedule replies after a certain delay.
+/// Delays are randomly chosen following a Pareto Distribution.
+/// 80% of the delay will be below 20% of MAX_LATENCY_MICROS
+pub struct LatencyInjector {
+    rng: rand_pcg::Pcg64,
+    distr: rand_distr::Pareto<f64>,
+    tx: Sender<(Reply, Duration)>,
+}
+
+impl LatencyInjector {
+    pub fn new(seed: u64) -> Self {
+        let (tx, rx) = mpsc::channel();
+        let _ = std::thread::spawn(|| scheduler(rx));
+        Self {
+            rng: rand_pcg::Pcg64::seed_from_u64(seed),
+            distr: rand_distr::Pareto::new(1.0, 1.16).unwrap(),
+            tx,
+        }
+    }
+
+    pub fn schedule_reply(&mut self, reply: Reply) {
+        // Shift and scale, values above 100.0 (0.05%) are clipped to MAX_LATENCY_MICROS.
+        let f = f64::min((self.distr.sample(&mut self.rng) - 1.0) / 100.0, 1.0);
+        let micros = (f * MAX_LATENCY_MICROS as f64).round() as u64;
+        let delay = Duration::from_micros(micros);
+        self.tx.send((reply, delay)).unwrap();
+    }
+}
+
+/// Task used to execute every scheduled reply.
+fn scheduler(rx: Receiver<(Reply, Duration)>) {
+    let mut scheduled: VecDeque<(Reply, Instant)> = VecDeque::new();
+    loop {
+        let (_, deadline) = match scheduled.front() {
+            Some((reply, deadline)) => (reply, deadline),
+            None => {
+                // Nothing scheduled, wait for next reply.
+                match rx.recv() {
+                    Ok((reply, delay)) => {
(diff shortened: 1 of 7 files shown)


gabriele-0201:
* **#832** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/832?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/832?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

trickfs/src/lib.rs
```diff
@@ -406,6 +415,18 @@ impl Trick {
         }
         path
     }
+
+    /// Schedule the reply if `trigger_latency_injector` is on, otherwise reply directly.
+    fn schedule_reply(&mut self, reply: impl FnOnce() + Send + 'static) {
+        if !self
+            .trigger_latency_injector
```
pepyakin, line 422: Nit: can't we make this local, that is not atomic?

trickfs/src/latency.rs
```diff
@@ -0,0 +1,91 @@
...
+        let (_, deadline) = match scheduled.front() {
+            Some((reply, deadline)) => (reply, deadline),
+            None => {
+                // Nothing scheduled, wait for next reply.
+                match rx.recv() {
+                    Ok((reply, delay)) => {
+                        schedule_new_reply(&mut scheduled, reply, delay);
+                    }
+                    Err(_) => break,
+                };
```
pepyakin, line 58: Nit
```suggestion
                }
```

trickfs/src/latency.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    // Answer to all pending replies.
+    for (reply, _) in scheduled {
+        reply();
+    }
+}
+
+/// Insert the reply into the scheduled queue following an order by the deadlines.
+fn schedule_new_reply(scheduled: &mut VecDeque<(Reply, Instant)>, reply: Reply, delay: Duration) {
+    let new_deadline = std::time::Instant::now() + delay.clone();
```
pepyakin, line 84: Nit: not sure what makes it a *new* deadline

trickfs/src/latency.rs
```diff
@@ -0,0 +1,91 @@
...
+                    }
+                    Err(_) => break,
+                };
+                continue;
+            }
+        };
+
+        // Wait for a new reply to be scheduled or until we reach the deadline
+        // of the first reply in the queue.
+        let timeout = *deadline - std::time::Instant::now();
```
pepyakin, line 65: Is it possible that a negative duration will be created here (= panic)?

pepyakin (changes requested):
Blocking only because of potential panic.

trickfs/src/lib.rs
gabriele-0201: I don't think so. `trigger_latency_injector` is an `Arc<AtomicBool>` which is shared with `TrickHandle` in the following manner:
```rust
impl TrickHandle {
    pub fn set_trigger_latency_injector(&self, on: bool) {
        self.trigger_latency_injector
            .store(on, std::sync::atomic::Ordering::Relaxed);
    }
    //...
}
```
Thus `trigger_latency_injector` is being shared between the trickfs thread and the supervisor which should always be able to trigger it on/off.

It mimics what done by `trigger_enospc` which is also an `Arc<AtomicBool>`.

trickfs/src/latency.rs
gabriele-0201: Just the fact that it is being added to the `scheduled` deque, nothing more.

I will update it to:
``` rust
    let deadline = std::time::Instant::now() + delay.clone();
    // If two replies happen to have the same deadline, then they will be kept in FIFO order.
    let idx = match scheduled.binary_search_by_key(&deadline, |(_, d)| *d) {
        Ok(idx) => idx + 1,
        Err(idx) => idx,
    };
    scheduled.insert(idx, (reply, deadline));

```

trickfs/src/latency.rs
gabriele-0201: Initially I thought no because the implementation of `std::ops::Sub` for `std::time::Instant` states: `Returns the amount of time elapsed from another instant to this one, or zero duration if that instant is later than this one.` But your comment made me look again and I found a second portion of the doc which states: `Panics - Previous Rust versions panicked when other was later than self. Currently this method saturates. Future versions may reintroduce the panic in some circumstances. See [Monotonicity](https://doc.rust-lang.org/std/time/struct.Instant.html#monotonicity).`

I will update the code to use `checked_sub`!

pepyakin (approved):

PR #836, opened 2025-03-01
feat(torture): agent concurrent reads

gabriele-0201:
Add to the `CommitPayload` a vector of keys that need to be read from the session before applying the changeset.
Not only are the keys to read specified, but also the amount of concurrency with which the reads should be performed.

PR #834, opened 2025-03-01
feat(torture): spawn multiple parallel workloads

gabriele-0201:
Prepare the supervisor to be able to spawn multiple
workloads in parallel.
Logging all workloads to the same stdout was incomprehensible.
Now, each workload that is being orchestrated by the supervisor
and the relative agent will log into the same file at `workload_dir/log`.

NOTE: I also removed some TODOs from the logging module because 
I was not having any problem with ANSI colors.

PR #837, opened 2025-03-01
feat(torture): add resource allocator

gabriele-0201:
Currently, `ResourceAllocator` assigns random portions of disk space and memory to workloads, while ensuring that everything stays within well-defined boundaries.

`ResourceAllocator` makes it easy to customize the resource allocation algorithm by simply modifying the `alloc` method!

### DIFF
--- a/torture/src/supervisor/resource.rs
+++ b/torture/src/supervisor/resource.rs
@@ -1,0 +1,266 @@
+use std::{
+    io::Read,
+    os::unix::fs::MetadataExt,
+    path::{Path, PathBuf},
+};
+
+use rand::{Rng, SeedableRng};
+
+/// The maximum percentage of total disk space that torture will occupy.
+const MAX_DISK_OCCUPANCY_RATIO: f64 = 0.7;
+
+/// 1GiB is the minimum amount of disk space that can be assigned to a workload.
+const MIN_ASSIGNED_DISK: u64 = 1 * (1 << 30);
+
+/// The maximum percentage of total memory that torture will occupy.
+const MAX_MEMORY_OCCUPANCY_RATIO: f64 = 0.8;
+
+/// 100MiB is the minimum amount of memory that can be assigned to a workload.
+const MIN_ASSIGNED_MEMORY: u64 = 250 * (1 << 20);
+
+/// 5GiB is the maximum amount of memory that can be assigned to a workload
+/// which will use both disk and memory.
+const MAX_ASSIGNED_MEMORY: u64 = 5 * (1 << 30);
+
+/// 50GiB is the maximum amount of memory that can be assigned to a workload
+/// which will use only memory.
+const MAX_ASSIGNED_ONLY_MEMORY: u64 = 50 * (1 << 30);
+
+/// ResourceAllocator is used to split resources randomly across multiple workloads.
+///
+/// Resources are Memory and Disk space.
+pub struct ResourceAllocator {
+    rng: rand_pcg::Pcg64,
+    // (workload_id, assigned_disk, assigned_mem)
+    assigned: Vec<(u64, u64, u64)>,
+    max_disk_avail: u64,
+    max_memory_avail: u64,
+    total_assigned_disk: u64,
+    total_assigned_memory: u64,
+}
+
+impl ResourceAllocator {
+    /// Creates a `ResourceAllocator` given the available disk space at the path of the workdir,
+    /// where all workload data will be saved.
+    pub fn new(workdir_path: PathBuf, seed: u64) -> anyhow::Result<Self> {
+        let (avail_disk, total_disk) = disk_info(&workdir_path);
+        let occupied_disk = total_disk - avail_disk;
+        let max_disk_occupancy = (total_disk as f64 * MAX_DISK_OCCUPANCY_RATIO) as u64;
+        let Some(max_disk_avail) = max_disk_occupancy.checked_sub(occupied_disk) else {
+            anyhow::bail!("Free disk space is less than what was expected to be occupied at most");
+        };
+
+        let (avail_memory, total_memory) = mem_info();
+        let occupied_memory = total_memory - avail_memory;
+        let max_memory_occupancy = (total_memory as f64 * MAX_MEMORY_OCCUPANCY_RATIO) as u64;
+        let Some(max_memory_avail) = max_memory_occupancy.checked_sub(occupied_memory) else {
+            anyhow::bail!("Free memory is less than what was expected to be occupied at most");
+        };
+
+        Ok(Self {
+            rng: rand_pcg::Pcg64::seed_from_u64(seed),
+            assigned: vec![],
+            max_memory_avail,
+            max_disk_avail,
+            total_assigned_disk: 0,
+            total_assigned_memory: 0,
+        })
+    }
+
+    /// It assigns portions of disk space and memory randomly to the specified `workload_id`.
+    pub fn alloc(&mut self, workload_id: u64) -> Result<(), ResourceExhaustion> {
+        // Assign disk space
+        let avail_disk = self.max_disk_avail - self.total_assigned_disk;
+        if avail_disk <= MIN_ASSIGNED_DISK {
+            return Err(ResourceExhaustion::Disk);
+        }
+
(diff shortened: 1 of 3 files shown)


gabriele-0201:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/837?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#837** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/837?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/837?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#836** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/836?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#835** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/835?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#834** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/834?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#833** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/833?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#832** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/832?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

torture/src/supervisor/resource.rs
```diff
@@ -0,0 +1,218 @@
...
+        .split('\n')
+        .find(|entry| entry.starts_with(name))
+        .and_then(|entry| entry.split(':').nth(1))
+        .and_then(|value| value.trim_start().split(' ').nth(0))
+        .and_then(|value| value.parse::<u64>().ok())
+        .unwrap();
+    value * 1024
+}
+
+// Returns the total number of bytes avaialbe on the disk where the `workdir` is located.
```
pepyakin, line 196:
```suggestion
// Returns the total number of bytes available on the disk where the `workdir` is located.
```

torture/src/supervisor/resource.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        if avail_disk < MIN_ASSIGNED_DISK_SPACE {
+            anyhow::bail!("Not enough disk space to instantiate a new workload");
+        }
+
+        let assigned_disk = self.rng.gen_range(MIN_ASSIGNED_DISK_SPACE..avail_disk);
+
+        // Assign memory
+        let avail_mem = avail_mem();
+        let avil_mem_ratio = avail_mem as f64 / total_mem() as f64;
```
pepyakin, line 66: I am concerned that the code fetches memory data in two shots which is a gateway for a race condition. While we are here maybe it would be good to guard against div by zero here as well

torture/src/supervisor/resource.rs
```diff
@@ -0,0 +1,218 @@
...
+/// 100MiB is the minimum amount of memory that can be assigned to a workload.
+const MIN_ASSIGNED_MEMORY: u64 = 250 * (1 << 20);
+
+/// ResourceAllocator is used to split resources randomly across multiple workloads.
+///
+/// Resources are Disk space and Memory available.
+pub struct ResourceAllocator {
+    rng: rand_pcg::Pcg64,
+    // (workload_id, assigned_disk, assigned_mem)
+    assigned: Vec<(u64, u64, u64)>,
```
pepyakin, line 30: why not introduce a nominal type?

torture/src/supervisor/resource.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub fn assigned_disk(&mut self, workload_id: u64) -> u64 {
+        self.assigned
+            .iter()
+            .find(|(id, ..)| *id == workload_id)
+            .map(|(_id, assigned_disk, _)| *assigned_disk)
+            .unwrap()
+    }
+}
+
+// Return the amount of physical memory utilized by the specified pid.
```
pepyakin, line 159: Why not rustdoc?

```suggestion
/// Return the amount of physical memory utilized by the specified pid.
```

torture/src/supervisor/resource.rs
```diff
@@ -0,0 +1,218 @@
...
+pub fn process_memory_occupied(pid: u32) -> u64 {
+    let mut statm_file = std::fs::File::open(format!("/proc/{}/statm", pid)).unwrap();
```
pepyakin, line 161: Wouldn't this panic?

pepyakin (changes requested):
Would be good to resolve the issue with the panic

pepyakin:
Not sure if blocking, since the code is not being used.

torture/src/supervisor/resource.rs
gabriele-0201: > the code fetches memory data in two shots
tbf there is no real reason why `avail_mem` is separate from `total_mem`, I will update the code to `mem_info` which returns avail and total memory!

Guarding against div by zero is definitely a good thing but also a better safe than sorry approach here because I'm not sure how the returned total memory could be zero without the system being broken.

torture/src/supervisor/resource.rs
gabriele-0201: Not really a fan of those if they would have been used just once or twice internally.

I will give it a try and see if they make the code look nicer!

torture/src/supervisor/resource.rs
gabriele-0201: Not sure, probably just a mistake. I will update rustdoc for all functions!

torture/src/supervisor/resource.rs
gabriele-0201: Are you referring to the fact that this function could panic and it is not stated in the docs? Or that there is a panic ready to be triggered somewhere that I'm not able to see?

If it is the first thing, then the comment is now only present in `is_exceeding_resources`. I will also copy it to the internal function

torture/src/supervisor/resource.rs
```diff
@@ -0,0 +1,269 @@
...
+/// Panics if pid does not refer to an ongoing process.
+pub fn process_memory_occupied(pid: u32) -> u64 {
+    let file_path = format!("/proc/{}/statm", pid);
+    let mut statm_file = std::fs::File::open(&file_path).expect(
+        format!(
+            "Getting process memory information failed, {} does not exist",
+            file_path
+        )
+        .as_str(),
+    );
```
pepyakin, line 221: I guess my problem with this is that there is not enough assurance that the process is alive. What guarantees that exactly? I cannot see what. The process controlled by the `Child` object can die under its nose, and thus `/proc/<pid>/statm` will not exist.

torture/src/supervisor/resource.rs
```diff
@@ -0,0 +1,269 @@
...
+        let mut avail_memory = self.max_memory_avail - self.total_assigned_memory;
+        if avail_memory <= MIN_ASSIGNED_MEMORY {
+            return Err(ResourceExhaustion::Memory);
+        }
+
+        avail_memory = std::cmp::min(avail_memory, MAX_ASSIGNED_ONLY_MEMORY);
+        let assigned_mem = self.rng.gen_range(MIN_ASSIGNED_MEMORY..avail_memory);
+
+        self.assigned.push((workload_id, 0, assigned_mem));
+        Ok(())
```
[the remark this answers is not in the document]
gabriele-0201: This has already been fixed in a [follow up](https://github.com/thrumdev/nomt/pull/847/files#diff-3484ff5f7799ba02bff361a98a98697745811c5a8f03b854ee2ac6103a25f42cR119)!

torture/src/supervisor/resource.rs
gabriele-0201: Yes, you are totally right. I will update the code to avoid panicking, and if the expected file is not in the right place, I will simply deallocate the resources!

pepyakin (approved):

PR #849, opened 2025-03-24
Suggestion: Allow to open nomt in a new empty directory

citizen-stig:
it might be useful to open nomt in temporary folder. 

Of course it can be solved on user end by appending "nomt" as path inside newly created temp directory, but I think it is slightly better dev-ex


pepyakin (approved):
Yeah, that's reasonable. Thanks!

PR #853, opened 2025-03-27
chore: remove borsh version limit

### LINKED ISSUE #850: Remove borsh version limit
I see that borsh version has been limited in this commit: 
https://github.com/thrumdev/nomt/commit/44c47df9e469872702b17ad399dfaf396ef67e53

Is it possible to lift this limit?

It makes it impossible to use nomt and [risc0-zkvm](https://github.com/risc0/risc0/blob/ae90eaefa1da8b222955f0f74b4995d74bd1dbb8/risc0/zkvm/Cargo.toml#L33) in the same project

citizen-stig:
Issue described in #850

Resolves #850 

This is more of exploratory PR to understand what was original reason to lock it, and if this problem is still present now

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -219,26 +219,25 @@ dependencies = [
 
 [[package]]
 name = "borsh"
-version = "1.4.0"
+version = "1.5.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0901fc8eb0aca4c83be0106d6f2db17d86a08dfc2c25f0e84464bf381158add6"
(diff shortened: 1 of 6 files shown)


gabriele-0201 (changes requested):
Given the problem you explained in #850, I don't see why we should have two different `borsh` versions between `nomt` and `nomt_core`. I tried compiling with both versions changed to the latest `1.5.7`, and it works, so why not stick to the latest instead of having weird dependency requirements?

citizen-stig:
I moved dependency declaration to workspace level and set it to the latest 1.5.7

pepyakin (approved):

gabriele-0201 (approved):

PR #858, opened 2025-04-04
fix(torture): existing key generation

gabriele-0201:


gabriele-0201:
* **#858** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/858?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/858?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

pepyakin (approved):

torture/src/supervisor/workload.rs
```diff
@@ -955,35 +955,52 @@ impl Workload {
...
             let mut key = [0; 32];
             rng.fill_bytes(&mut key);
-            let mut start_key = None;
+            let start_key = key;
+            let mut restart = false;
             // Starting from a random key, perform a circular linear search looking
             // for an unused key to delete. This is never called if the committed state is empty,
             // but we need to check that not all committed keys are already used.
             loop {
+                if restart && key > start_key {
```
pepyakin, line 964: I am not entirely sure about this condition,

gabriele-0201: The goal is to stop only when we have restarted the circular linear search and when the key we reach is greater than the one we started from. If we do only one of those, we would end up ending the loop on the second iteration (since the second key will be bigger than the first) or we would not scan all the keys from the beginning to the key we started from.

PR #859, opened 2025-04-04
feat: count merkle leaves in ht pages

gabriele-0201:
Add a feature that counts the number of leaves
in each subtree on a bitbox page.

This pr also introduces two new metrics to keep track of how many pages were elided and how many were actually created or modified.

### DIFF
--- a/nomt/src/merkle/page_walker.rs
+++ b/nomt/src/merkle/page_walker.rs
@@ -54,9 +54,16 @@ use nomt_core::{
 
 use crate::{
     merkle::BucketInfo,
+    metrics::Metrics,
     page_cache::{Page, PageMut},
     page_diff::PageDiff,
 };
+
+/// Threshold representing the number of leaves required to be present in the two
+/// subtrees contained in a page to be stored on disk.
+/// If this threshold is not exceeded, the page will not be stored on disk
+/// and will be constructed on the fly when needed.
+const PAGE_ELISION_THRESHOLD: u64 = 20;
 
 /// The output of the page walker.
 pub enum Output {
@@ -81,6 +88,10 @@ pub struct UpdatedPage {
     pub diff: PageDiff,
     /// The bucket info associated with the page.
     pub bucket_info: BucketInfo,
+    /// If Some contains a counter of all the leaves present in the two subtrees
+    /// of the page. If the counter exceeds [`PAGE_ELISION_THRESHOLD`], we don't
+    /// need to keep it updated, and it becomes None.
+    pub leaves_counter: Option<u64>,
 }
 
 /// A set of pages that the page walker draws upon.
@@ -110,13 +121,14 @@ pub struct PageWalker<H> {
     sibling_stack: Vec<(Node, usize)>,
     prev_node: Option<Node>, // the node at `self.position` which was replaced in a previous call
 
+    metrics: Metrics,
     _marker: std::marker::PhantomData<H>,
 }
 
 impl<H: NodeHasher> PageWalker<H> {
     /// Create a new [`PageWalker`], with an optional parent page for constraining operations
     /// to a subsection of the page tree.
-    pub fn new(root: Node, parent_page: Option<PageId>) -> Self {
+    pub fn new(root: Node, parent_page: Option<PageId>, metrics: Metrics) -> Self {
         PageWalker {
             last_position: None,
             position: TriePosition::new(),
@@ -127,6 +139,7 @@ impl<H: NodeHasher> PageWalker<H> {
             stack: Vec::new(),
             sibling_stack: Vec::new(),
             prev_node: None,
+            metrics,
             _marker: std::marker::PhantomData,
         }
     }
@@ -163,7 +176,6 @@ impl<H: NodeHasher> PageWalker<H> {
     /// Advance to a given trie position and place the given node at that position.
     ///
     /// It is the responsibility of the user to ensure that:
-    ///   - if this is a leaf node, the leaf data have been written to the two child positions.
     ///   - if this is an internal node, the two child positions hashed together create this node.
     ///   - if this is a terminal node, then nothing exists in the two child positions.
     ///
@@ -308,8 +320,13 @@ impl<H: NodeHasher> PageWalker<H> {
     fn up(&mut self) {
         if self.position.depth_in_page() == 1 {
             // UNWRAP: we never move up beyond the root / parent page.
-            let stack_item = self.stack.pop().unwrap();
-            self.updated_pages.push(stack_item);
+            let stack_top = self.stack.pop().unwrap();
+            handle_elision_threshold::<H>(
+                stack_top,
+                &mut self.stack,
+                &mut self.updated_pages,
+                &mut self.metrics,
+            );
         }
         self.position.up(1);
     }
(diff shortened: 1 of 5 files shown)


gabriele-0201:
* **#861** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/861?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#860** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/860?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#859** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/859?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/859?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

nomt/src/metrics.rs
```diff
@@ -81,6 +94,12 @@ impl Metrics {
         if let Some(ref metrics) = self.metrics {
             println!("metrics");
 
+            let hashtable_pages = metrics.hashtable_pages.load(Ordering::Relaxed);
+            println!("  ht pages              {}", hashtable_pages);
```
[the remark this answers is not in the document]
rphmeier: yeah, could be more descriptive

nomt/src/metrics.rs
```diff
@@ -9,6 +9,9 @@ pub struct Metrics {
     metrics: Option<Arc<ActiveMetrics>>,
 }
 
+// Safety: Metrics is just a wrapper around ActiveMetrics, which is only made my `AtomicU64`.
```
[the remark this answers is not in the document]
rphmeier: Yeah, this looks unnecessary. Just delete?

nomt/src/merkle/page_walker.rs
```diff
@@ -54,10 +54,17 @@ use nomt_core::{
...
+    metrics::Metrics,
     page_cache::{Page, PageMut},
     page_diff::PageDiff,
 };
 
+/// Threshold representing the number of leaves required to be present in the two
+/// subtrees contained in a page to be stored on disk.
+/// If this threshold is not exceeded, the page will not be stored on disk
+/// and will be constructed on the fly when needed.
+const PAGE_ELISION_THRESHOLD: u64 = 10;
```
rphmeier, line 66: I would prefer if this is tunable. Also, I think that the value here can be higher without impacting perf.

nomt/src/merkle/page_walker.rs
```diff
@@ -333,6 +350,7 @@ impl<H: NodeHasher> PageWalker<H> {
                     page,
                     diff: PageDiff::default(),
                     bucket_info,
+                    leaves_counter: Some(0),
```
rphmeier, line 353: `Some(0)` doesn't seem accurate unless the page is fresh, at least according to the doc.

nomt/src/merkle/page_walker.rs
```diff
@@ -600,6 +626,83 @@ impl<H: NodeHasher> PageWalker<H> {
...
+}
+
+/// Count the number of leaves present *only* in the provided page,
+/// without jumping into child pages.
+fn count_leaves<H: NodeHasher>(page: &PageMut) -> u64 {
+    // A simpler linear scan cannot be done because the page could contain some garbage.
+    let mut counter = 0;
+
+    // We just need the node indexes within a page,
+    // so we treat this as the root page starting from the root positoin.
```
rphmeier, line 674: typo: position

nomt/src/merkle/page_walker.rs
```diff
@@ -600,6 +626,83 @@ impl<H: NodeHasher> PageWalker<H> {
...
+) {
+    if let Some(leaves_counter) = stack_top.leaves_counter {
+        let n_leaves = count_leaves::<H>(&stack_top.page);
+
+        if stack.is_empty() {
+            metrics.count(crate::metrics::Metric::HashTablePages);
+            updated_pages.push(stack_top);
+
+            return;
+        } else if n_leaves + leaves_counter < PAGE_ELISION_THRESHOLD {
```
rphmeier, line 643: this section feels odd to me, and may be wrong.

1. earlier, we initialize `leaves_counter` for any page to be equal to `Some(0)`.
2. then, when we pop a page from the stack we check its `leaves_counter` + the amount of self-contained leaves and determine whether to elide on that basis.
3. this implies that the `leaves_counter` for a page P is determined only by the sum of the leaves reachable in _accessed child pages_, not all child pages.


This seems broken to me. Let me provide the following example. If we have a page P with 2 children, P1 and P2. In P1 there are 2 leaves. In P2 there are 64. If we only load P and P1, then we would make the decision to elide P and P1, even though P obviously should not be elided.

nomt/src/metrics.rs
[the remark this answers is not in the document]
gabriele-0201: Wow, I was sure at some point I needed it to be able to send it across the merkle updater thread, but that's wrong! I will remove it!

nomt/src/merkle/page_walker.rs
gabriele-0201: Yes, how the number of leaves is managed will be revisited later. This code was only to test whether the counter correctly detected the number of pages that should have been elided!

nomt/src/merkle/page_walker.rs
```diff
@@ -600,6 +626,83 @@ impl<H: NodeHasher> PageWalker<H> {
...
+            metrics.count(crate::metrics::Metric::ElidedHashTablePages);
+
+            // UNWRAP: The stack has just been verified to not be empty
+            let new_stack_top = stack.last_mut().unwrap();
+            if let Some(ref mut upper_leaves_in_childs) = new_stack_top.leaves_counter {
+                *upper_leaves_in_childs += n_leaves + leaves_counter;
```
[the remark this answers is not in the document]
gabriele-0201: Yes, I was initially using 'childs,' then remembered that it's wrong, updated it everywhere, and missed it here :sweat_smile:

nomt/src/merkle/page_walker.rs
```diff
@@ -600,6 +626,83 @@ impl<H: NodeHasher> PageWalker<H> {
...
+            }
+
+            updated_pages.push(stack_top);
+            return;
+        }
+    }
+
+    if !stack.is_empty() {
+        // Propagates the exceeded page elision threshold.
+        let new_stack_top = stack.last_mut().unwrap();
```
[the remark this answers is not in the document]
gabriele-0201: You're right, good old rust and `if let` syntax

nomt/src/merkle/page_walker.rs
gabriele-0201: You are totally right! The counter, as it is in this PR, is broken. The goal of this PR was just to introduce the function `count_leaves`, and to avoid having unused things around, I left this code (which now I realize I should have removed when I fixed this problem). #863 contains the correct implementation.

In the correct implementation, the actual page origin is tracked, and leaves are counted only if it is `Reconstructed`. Otherwise, `leaves_counter` is None for each page fetched from the hash table and remains the same for all parents.

rphmeier: Yeah, in light of the full page elision PR I don't think my comments here are that relevant

pepyakin:
### Merge activity

* **Apr 14, 7:54 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thrumdev/nomt/859).

PR #864, opened 2025-04-09
Expose prev_root() getter for Session

citizen-stig:
Sovereign SDK operates on `Session` and `FinishedSession` and it needs access to prev_root to pass it to ZK guest.

I am happy to change location of it, but this method is needed for the integration.

Please let me know if there's any issue of exposing it.

### DIFF
--- a/nomt/src/lib.rs
+++ b/nomt/src/lib.rs
@@ -572,7 +572,7 @@ impl<T: HashAlgorithm> Session<T> {
         self.store.load_value(path)
     }
 
-    /// Prev root
+    /// Previous root
     pub fn prev_root(&self) -> Root {
         self.prev_root
     }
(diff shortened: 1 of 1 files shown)


gabriele-0201:
I see how accessing the `prev_root` can be required, and I'm totally okay with adding this helper! I would probably move it into `FinishedSession` rather than within `Session` because it seems more logical to access the `prev_root` after a set of changes has been applied.

pepyakin:
This one touches a public API and I would rather err on the safe side. Although on the glance it seems to be reasonable. Can you give a bit more detail on how exactly it is needed?

pepyakin (approved):
That's reasonable and I think we could add it to `FinishedSession` as well.

PR #865, opened 2025-04-09
Use `digest` crate for Blanket implementation of `BinaryHash`

citizen-stig:
In Sovereign SDK we use `digest` as generic way to reprsent different hashers. So SDK itself does not depend on particular hasher implementation:

```
pub trait CryptoSpec {
    type Hasher: digest::Digest<OutputSize = digest::typenum::U32>;
    // The rest is omitted
}
```

This PR adds blanket implementation of `BinaryHash` for all Digest with matching output size.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1096,6 +1096,7 @@ dependencies = [
  "bitvec",
  "blake3",
  "borsh",
+ "digest",
  "hex",
  "ruint",
  "serde",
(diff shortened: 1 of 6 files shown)


pepyakin (approved):
Seems reasonable, we depend on `digest` via `sha2` indirectly, so might as well provide this blanket impl.

PR #894, opened 2025-04-24
fix: empty multiproof creation no longer panics

rphmeier:
Closes #893

PR #870, opened 2025-04-10
fix: verify_update overflow fix

pepyakin:
It may happen that there are no shared bits because the paths are empty.
In that case, there is going to be an underflow.

UPD: turned out my initial fix is not sufficient, so posting here only the failing test with an intent to update it later.

UPD: Solved this by moving the preconditions check into a single pass before running any logic.

PR #895, opened 2025-04-25
feat: introduce a verify_multi_proof_update

rphmeier:
This turned out more involved than I expected.

The `VerifiedMultiProof` has been extended with a `Vec<VerifiedBisection>` that keeps tracks of all the bisections and upper sibling ranges. This is used to recover all needed siblings.

The core of this operates quite similarly to the existing `verify_update`:
  1. loop through all terminals (ascending)
  2. apply any needed changes to terminals based on `ops` as we go
  
most of the complexity is in gathering the necessary siblings and making sure they are available. for that I've introduced a helper struct `CommonSiblings`.
  
I added a few tests, including one which explicitly loads the bisection logic.
  
Runtime is linear in the number of multiproofs, assuming an average depth per branch. 

This algorithm, as written, does reproduce some of the work from `verify_update`.

### DIFF
--- a/core/src/proof/multi_proof.rs
+++ b/core/src/proof/multi_proof.rs
@@ -8,14 +8,14 @@ use crate::{
         path_proof::{hash_path, shared_bits},
         KeyOutOfScope, PathProof, PathProofTerminal,
     },
-    trie::{InternalData, KeyPath, LeafData, Node, TERMINATOR},
+    trie::{InternalData, KeyPath, LeafData, Node, NodeKind, ValueHash, TERMINATOR},
 };
 
 #[cfg(not(feature = "std"))]
 use alloc::{vec, vec::Vec};
 
 use bitvec::prelude::*;
-use core::cmp::Ordering;
+use core::{cmp::Ordering, ops::Range};
 
 /// This struct includes the terminal node and its depth
 #[derive(Debug, Clone)]
@@ -277,12 +277,23 @@ pub enum MultiProofVerificationError {
 struct VerifiedMultiPath {
     terminal: PathProofTerminal,
     depth: usize,
+    unique_siblings: Range<usize>,
+}
+
+// indicates a bisection which started at a given depth and covers these common siblings.
+#[derive(Debug, Clone)]
+struct VerifiedBisection {
+    start_depth: usize,
+    common_siblings: Range<usize>,
 }
 
 /// A verified multi-proof.
 #[derive(Debug, Clone)]
 pub struct VerifiedMultiProof {
     inner: Vec<VerifiedMultiPath>,
+    bisections: Vec<VerifiedBisection>,
+    siblings: Vec<Node>,
+    root: Node,
 }
 
 impl VerifiedMultiProof {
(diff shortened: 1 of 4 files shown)


rphmeier:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/895?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#895** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/895?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/895?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#894** <a href="https://app.graphite.dev/github/pr/thrumdev/nomt/894?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

core/src/proof/multi_proof.rs
```diff
@@ -500,15 +552,338 @@ fn verify_range<H: NodeHasher>(
...
+        .by_vals()
+        .rev()
+        .take(up_layers)
+    {
+        let sibling = if pending_siblings.last().map_or(false, |p| p.1 == cur_layer) {
+            // is this even possible? maybe not. but being extra cautious...
+            let _ = common_siblings.pop_if_at_depth(cur_layer);
+            // UNWRAP: guaranteed to exist.
+            pending_siblings.pop().unwrap().0
+        } else {
```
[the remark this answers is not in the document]
rphmeier: too obvious IMO

core/src/proof/multi_proof.rs
```diff
@@ -500,15 +552,338 @@ fn verify_range<H: NodeHasher>(
...
+    let skip = terminal.depth;
+
+    let up_layers = if let Some(next_terminal) = next_terminal {
+        let n = shared_bits(terminal.terminal.path(), next_terminal.terminal.path());
+        // n always < skip
+        // we want to end at layer n + 1
+        skip - (n + 1)
+    } else {
+        skip // go to root
+    };
```
pepyakin, line 817: I think this exhibits exactly the same problem as I fixed in https://github.com/thrumdev/nomt/commit/6381c5412823a9fa9ae2c3bd55f7afd3254ce025

I asked chatgpt to come up with a test and here is a result that panics with underflow

https://github.com/thrumdev/nomt/commit/af6d1199431a2e693431d2fdeb64bbe68124a5d4

pepyakin (changes requested):

core/src/proof/multi_proof.rs
rphmeier: I see where you're coming from, and this test does indeed fail, but...

A premise of this test is that you can do

```rust
let bits_prefix = bitvec![u8, Msb0; 1, 0, 1, 0]; // length 4
let bits_longer = bitvec![u8, Msb0; 1, 0, 1, 0, 1, 1]; // length 6 (prefix + 2 bits)

let vmp_prefix = VerifiedMultiPath {
    terminal: PathProofTerminal::Leaf(leaf_prefix.clone()),
    depth: bits_prefix.len(),
    unique_siblings: 0..0, // Minimal example
};

let vmp_longer = VerifiedMultiPath {
    terminal: PathProofTerminal::Leaf(leaf_longer.clone()),
    depth: bits_longer.len(),
    unique_siblings: 0..0, // Minimal example
};
```

This is telling the `verify_update` function that there is a leaf at `0b1010` and a leaf at `0b101011`, which is impossible.

The same premise applies for #870 frankly.

rphmeier: Anyway the fix is small so might as well. But there might be other garbage inputs that also cause panics and I don't know if we can catch all of them.

PR #901, opened 2025-05-25
Deletions with overlays reproducible bug

omerb4n:


omerb4n:
To see reproduction run one of these commands in the benchtop dir:
* cargo run --bin unreachable --release
* cargo run --bin left_right_assertion --release

PR #936, opened 2026-05-12
Adding more property tests and fixing edge case of dropped operation

citizen-stig:
As follow up of https://github.com/thrumdev/nomt/pull/931

Includes a fix for:

`RangeUpdater::new` computed `range_end` with `binary_search_by_key(...).unwrap_or_else(|i| i)`.

When a key in `read_write` exactly equals the worker's `key_range_end` (the max key path of the shard's region), binary_search_by_key returns `Ok(i)` and `unwrap_or_else` returns `i` instead of the required exclusive end `i+1`.

The matching key is dropped from the worker's range, so the seek is never pushed, no `RootPagePending::SubTrie` is emitted, and the second commit's root just inherits the first commit's leaf-root unchanged.

Tracing the failing case showed `range_end=0` for the second commit (read_write contains only `KEY_B = [0xFF; 32]`, and `key_range_end = [0xFF; 32]` for the single shard's universe region).

Note from @rphmeier 

> i will note that this situation would be extremely unlikely to occur with genuinely hashed keys due to the way we shard the key-space among workers, as it essentially requires a hash collision. but good future proofing.

## Notable changes

the old account_path filled only bytes 0..16 (`for i in 0..4 { path[i*4..][..4]... }`); the new one in nomt-test-utils/src/lib.rs:599-611 uses c`hunks_exact_mut(4)` and **fills all 32 bytes**. This is the real reason add_remove_1000's expected_roots changed — not a "fixture refresh" but a different key distribution. The new behavior is arguably better (more entropy), but it silently changes any test that depended on the old layout

### DIFF
--- a/nomt/src/merkle/worker.rs
+++ b/nomt/src/merkle/worker.rs
@@ -248,7 +248,7 @@ fn update<H: HashAlgorithm>(
             } => {
                 let ops = subtrie_ops(&shared.read_write[range_start..range_end]);
                 let ops = nomt_core::update::leaf_ops_spliced(prev_terminal, &ops);
-                root_page_updater.advance_and_replace(&page_set, trie_pos.clone(), ops.clone());
+                root_page_updater.advance_and_replace(&page_set, trie_pos.clone(), ops);
             }
         }
     }
@@ -297,10 +297,10 @@ impl<H: HashAlgorithm> RangeUpdater<H> {
             .binary_search_by_key(&key_range_start, |x| x.0)
             .unwrap_or_else(|i| i);
 
+        // `key_range_end` is inclusive for the worker region, so compute an upper bound here.
         let range_end = shared
             .read_write
-            .binary_search_by_key(&key_range_end, |x| x.0)
-            .unwrap_or_else(|i| i);
+            .partition_point(|(key, _)| *key <= key_range_end);
 
         RangeUpdater {
             shared,
--- a/nomt/tests/add_remove.rs
+++ b/nomt/tests/add_remove.rs
@@ -9,18 +9,19 @@ fn add_remove_1000() {
     let mut accounts = 0;
     let mut t = Test::new("add_remove");
 
+    // These fixtures track the current `common::account_path` distribution.
     let expected_roots = [
         hex!("0000000000000000000000000000000000000000000000000000000000000000"),
-        hex!("4a7a6fe118037086a49ff10484f4d80b0a9f31f1060eeb1c9f0162634604b0d9"),
-        hex!("7d5b013105d7b835225256f2233a458e1a158a53d20e0d3834886df89a26c27b"),
-        hex!("1a290e07bcacfb58ddcd0b9da348c740ca1bf87b05ed96752a1503ed7c187b69"),
-        hex!("5e9abfee6d927b084fed3e1306bbe65f0880d0b7de12522c38813014927f1336"),
-        hex!("57b39e06b2ee98dccd882033eb4136f5376699128b421c83bdc7c6ca96168938"),
-        hex!("7fd75809ef0e2133102eb5e31e47cb577149dcaebb42cddeb2fd6754256b365f"),
-        hex!("7c00cb11ec8262385078613e7b7977e50b0751f8cb2384fdccc048eea02acb63"),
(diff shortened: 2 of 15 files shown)


nomt/tests/add_remove.rs
```diff
@@ -9,18 +9,19 @@ fn add_remove_1000() {
...
-        hex!("7d5b013105d7b835225256f2233a458e1a158a53d20e0d3834886df89a26c27b"),
-        hex!("1a290e07bcacfb58ddcd0b9da348c740ca1bf87b05ed96752a1503ed7c187b69"),
-        hex!("5e9abfee6d927b084fed3e1306bbe65f0880d0b7de12522c38813014927f1336"),
-        hex!("57b39e06b2ee98dccd882033eb4136f5376699128b421c83bdc7c6ca96168938"),
-        hex!("7fd75809ef0e2133102eb5e31e47cb577149dcaebb42cddeb2fd6754256b365f"),
-        hex!("7c00cb11ec8262385078613e7b7977e50b0751f8cb2384fdccc048eea02acb63"),
-        hex!("516d6911c3b0a36c9227922ca0273a4aee44886201bd186f7ee7e538a769eaa5"),
-        hex!("381b24719ff91b13d36cf0dd7622f391f4a461452ed7547a46a992ee4a4025aa"),
-        hex!("207793e2ce76c1feb68c7259f883229f985706c8cc2fcf99f481b622a54ba375"),
+        hex!("18a457ed03d9d28f8bf84aa05dc3c8005e35564e91d05bf6131094d4f3398528"),
```
rphmeier, line 15: this smells like a breaking change?

nomt/src/merkle/worker.rs
```diff
@@ -297,10 +297,10 @@ impl<H: HashAlgorithm> RangeUpdater<H> {
             .binary_search_by_key(&key_range_start, |x| x.0)
             .unwrap_or_else(|i| i);
 
+        // `key_range_end` is inclusive for the worker region, so compute an upper bound here.
         let range_end = shared
             .read_write
-            .binary_search_by_key(&key_range_end, |x| x.0)
-            .unwrap_or_else(|i| i);
+            .partition_point(|(key, _)| *key <= key_range_end);
```
rphmeier, line 303: i see, if `key_range_end` is actually in the set of ops we would drop it from the range silently. seems like a good fix.

i will note that this situation would be extremely unlikely to occur with genuinely hashed keys due to the way we shard the key-space among workers, as it essentially requires a hash collision. but good future proofing.

nomt/tests/add_remove.rs
citizen-stig: Yeah...
Byt as far I understand, this is for the test fixture, in reality this will unlikely produce different root hash.

rphmeier: Can you explain why the root hashes changed here and would not in production?

citizen-stig: I've mentioned it in PR description. This PR changed how `KeyPath` is derived from id

Before it has been filling only 16 bytes:

```rust
    for i in 0..4 {
        // max(i) == 3, path[12..][..4]
        path[i * 4..][..4].copy_from_slice(&rng.next_u32().to_le_bytes());
    }
```

https://github.com/thrumdev/nomt/blob/7ef155bc98b4fd04bdf7ba244c2be013bf56a4da/nomt/tests/common/mod.rs#L13-L24

And now it fills all 32 bytes:
https://github.com/thrumdev/nomt/blob/ed8ad2c2e4955672d8b490ad7b3d08ae8f67edde/nomt-test-utils/src/lib.rs#L342-L353

citizen-stig: If we want to play safe, we can do following:

1. Merge: https://github.com/thrumdev/nomt/pull/939
2. Do a patch release `1.0.3 -> 1.0.4`
3. Merge this PR
4. Do a breaking change release, `1.0.4 -> 1.1.0` or even more `2.0` to be more aligned with semver

rphmeier (approved):

nomt/tests/add_remove.rs
rphmeier: No need for that, I think this is OK. i see now that it's just because the test environment changing.

PR #941, opened 2026-06-05
Grow hashtable

theodorebugnet:
Adds a mostly vibe-coded implementation of growing the bucket storage on-disk. This way, if the buckets start to run out, the database can be grown in-place rather than needing to be rebuilt from scratch.

A large part of the line count is integrations into the `fuzz` and `torture` frameworks.

This has been empirically tested on rollups backed by NOMT storage, but most of the implementation was written by codex, so cleanup or comments about the approach are welcome!

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -58,3 +58,10 @@ path = "fuzz_targets/reconstruct_key.rs"
 test = false
 doc = false
 bench = false
+
+[[bin]]
+name = "grow_hashtable"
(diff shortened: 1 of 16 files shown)


rphmeier (commented):
Looks good to me, it's pretty simple.

You will get faster performance by refactoring it to use the async I/O facilities rather than the current approach, which synchronously reads each page individually and then writes it. By a quite significant factor, maybe as much as 10x. Fine as a follow-up

